PR #2, opened 2017-09-25
INT-355_test_iq_server_dockerfile

fernau:
JIRA: https://issues.sonatype.org/browse/INT-355

I would prefer serverspec.org for testing docker images. It is essentially the same what we do for chef, but I found kitchen not to work very well with docker.

- serverspec does not need SSH in the docker container
- serverspec does not hijack the 'CMD' command to run its tests
- serverspec runs the tests inside the container

### DIFF
--- a/spec/Dockerfile_spec.rb
+++ b/spec/Dockerfile_spec.rb
@@ -25,13 +25,12 @@ describe 'Dockerfile' do
     set :docker_image, @image.id
   end
 
-  it 'should generate a valid solo.json file' do
-    solo = JSON.parse(File.read('/var/chef/solo.json'))
-    expect(solo['run_list']).to eq([ 'recipe[nexus-iq-server::docker]' ])
+  it 'should remove solo.json during cleanup' do
+    expect(File).not_to exist('/var/chef/solo.json')
   end
 
   it 'should not have a chef package installed' do
-    expect(package("chef")).not_to be_installed
+    expect(package('chef')).not_to be_installed
   end
 
   it 'should have a user named nexus' do
--- a/spec/Dockerfile_spec.rb
+++ b/spec/Dockerfile_spec.rb
@@ -1,9 +1,21 @@
-# spec/Dockerfile_spec.rb
+# Copyright (c) 2017-present Sonatype, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


spec/Dockerfile_spec.rb
```diff
@@ -0,0 +1,33 @@
+# spec/Dockerfile_spec.rb
```
whyjustin: License header

spec/Dockerfile_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    Docker.options[:read_timeout] = 900
+    @image = Docker::Image.build_from_dir('.')
+
+    set :os, family: :redhat
+    set :backend, :docker
+    set :docker_image, @image.id
+  end
+
+  it "solo.json file is valid" do
+    solo = JSON.parse(File.read('/var/chef/solo.json'))
```
whyjustin: Single or double quotes consistently

spec/Dockerfile_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    expect(solo['run_list']).to eq([ 'recipe[nexus-iq-server::docker]' ])
+  end
+
+  it "chef package is installed" do
+    expect(package("chef")).to be_installed
+  end
+
+  it "nexus java process is running" do
+    expect(process('java')).to be_running
+    expect(process('java')[:user]).to eq('nexus')
```
whyjustin: Can we verify more with the process args? Or just the user?

whyjustin (changes requested):

spec/Dockerfile_spec.rb
fernau: we can verify most things with process. here is the link: http://serverspec.org/resource_types.html#process

spec/Dockerfile_spec.rb
fernau: added license header in 0007ead

spec/Dockerfile_spec.rb
fernau: fixed inconsistent use of quotes in 0007ead

spec/Dockerfile_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    Docker.options[:read_timeout] = 900
+    @image = Docker::Image.build_from_dir('.')
+
+    set :os, family: :redhat
+    set :backend, :docker
+    set :docker_image, @image.id
+  end
+
+  it 'should generate a valid solo.json file' do
+    solo = JSON.parse(File.read('/var/chef/solo.json'))
```
bigspotteddog: This file will not exist when after the cleanup happens. I added a change in #3 to address this.

bigspotteddog (approved):

whyjustin (approved):

PR #5, opened 2017-10-21
INT - Modify documentation and args

bigspotteddog:
* Expose the version and hash as build time args
* Correct volume documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,8 @@ LABEL vendor=Sonatype \
   com.sonatype.name="Nexus IQ Server image"
 
 # Optional parameters. Uncomment to override default:
-ENV IQ_SERVER_VERSION=1.36.0-01
-ENV IQ_SERVER_SHA256=a1a5d127e2deaf2b26ed4e605802986b99ea317c8a821059c918783517728d6c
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -21,8 +21,8 @@ LABEL vendor=Sonatype \
   com.sonatype.name="Nexus IQ Server image"
 
 # Optional parameters. Uncomment to override default:
-ENV IQ_SERVER_VERSION=1.36.0-01
-ENV IQ_SERVER_SHA256=a1a5d127e2deaf2b26ed4e605802986b99ea317c8a821059c918783517728d6c
+ARG IQ_SERVER_VERSION=1.36.0-01
+ARG IQ_SERVER_SHA256=a1a5d127e2deaf2b26ed4e605802986b99ea317c8a821059c918783517728d6c
```
fernau, line 25: Should JAVA_* below be ENV or ARG ?

fernau (approved):
well spotted 👍

Dockerfile
bigspotteddog: I thought about that too, but went with what docker-nexus3 did. They left them as ENV.

PR #16, opened 2018-02-02
Update IQ Server to 1.43.0-01

fernau:
https://issues.sonatype.org/browse/INT-444


whyjustin:
This was handled by #17.

PR #22, opened 2018-08-16
Provide clearer examples on server config

peterlynch:
Users of the official docker image need examples on how to customize the server configuration.

Here I try to add these examples and clarify where the config.yml is actually loaded from inside the docker image.


whyjustin (approved):

whyjustin:
Thanks @peterlynch

PR #26, opened 2019-01-09
fix toc typo

peterlynch:


whyjustin (approved):

whyjustin (changes requested):
This is actually the wrong syntax for hyperlinking in markdown. The lack of `.` is expected.

whyjustin (approved):
🤦‍♂️ Having internal malfunctions, this is OK.

PR #30, opened 2019-04-18
(feat): Add in Dockerfile iq-server running off of openjdk:alpine

timlevett:
* Creates alpine directory for alpine version of iq-server
* Adds `Dockerfile` and does all the steps chef commands does
* Adds stock `config.yml` that is generated from chef config (fetched with `docker run -it --rm sonatype/nexus-iq-server cat /etc/nexus-iq-server/config.yml`)

Not covered
* Testing
* checksum checking of download
* Building/tagging/deploying

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -1,0 +1,50 @@
+FROM openjdk:8u201-alpine
+
+LABEL maintainer="Sonatype <cloud-ops@sonatype.com>"
+
+LABEL vendor=Sonatype \
+  com.sonatype.license="Apache License, Version 2.0" \
+  com.sonatype.name="Nexus IQ Server image"
+
+# This is where we will store persistent data
+VOLUME /sonatype-work
+
+# Create needed directories
+RUN mkdir -p /opt/sonatype/nexus-iq-server /etc/nexus-iq-server /var/log/nexus-iq-server
+
+# Change working dir
+WORKDIR /opt/sonatype/nexus-iq-server
+
+# Add nexus user and group
+RUN adduser -s /bin/false -S nexus nexus
+RUN addgroup nexus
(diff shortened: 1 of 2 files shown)


alpine/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN adduser -s /bin/false -S nexus nexus
+RUN addgroup nexus
+RUN chown nexus:nexus /opt/sonatype/nexus-iq-server /etc/nexus-iq-server /var/log/nexus-iq-server
+
+# Change to nexus user
+USER nexus
+
+# Version to download
+ARG IQ_SERVER_VERSION=1.64.0-02
+ARG IQ_SERVER_SHA256=d79cae556c60982191a026d8407ca4c8b96f26876d088067fa4ae7ef1c39cb99
```
stangles, line 28: Just passing by and mainly taking a look at this out of my own interest, so if this doesn't make sense, feel free to ignore :)

Did you intend to use this to do an integrity check on the IQ Server binary that's fetched later on? If so, I don't see it used to compare against the hash of the binary.

timlevett: Ah, i meant to note that in the PR, thanks for pointing it out. I don't intend to merge this unless others have a need to have more control over the JDK.

alpine/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Fetch and configure IQ
+RUN mkdir /tmp/work && cd /tmp/work \
+&& wget https://download.sonatype.com/clm/server/nexus-iq-server-${IQ_SERVER_VERSION}-bundle.tar.gz \
```
tneer, line 32: So, I started thinking that this might be a case for using the `ADD` command (it isn't because it only extracts local files), but then I got to wondering why this is here and not part of https://github.com/sonatype/insight-brain/tree/master/nexus-iq-server

Not sure what the back story is, but seems like a good case for using one of the maven docker integrations to just build the image as part of the maven build.

PR #37, opened 2020-02-25
INT-2560 rhcc errors

adrianpowell:
fix some build errors

Jira: https://issues.sonatype.org/browse/INT-2560
Build: https://jenkins.ci.sonatype.dev/job/integrations/job/cloud/job/docker-nexus-iq-server-feature/job/INT-2560-RHCC-errors/


collinpeters (approved):
+1

Dockerfile.rhel
```diff
@@ -12,47 +12,45 @@
...
-    com.sonatype.license="Apache License, Version 2.0" \
+    automate your processes and achieve accelerated speed to delivery while also increasing product quality" \
   io.k8s.display-name="Nexus IQ Server" \
   io.openshift.expose-services="8071:8071" \
   io.openshift.tags="Sonatype,Nexus,IQ Server"
 
 # Optional parameters.
-ARG IQ_SERVER_VERSION=1.84.0-01
-ARG IQ_SERVER_SHA256=81e877f4e793e740cd0be38faaa35650448bfcb77aa30fd48d15babeb01a9a09
+ARG IQ_SERVER_VERSION=1.85.0-01
```
kellyrob99: seems like this should be used for the `version` and `release` values set above?

Dockerfile.rhel
```diff
@@ -12,47 +12,45 @@
...
   com.sonatype.license="Apache License, Version 2.0" \
   com.sonatype.name="Nexus IQ Server image" \
   run="docker run -d -p 8070:8070 -p 8071:8071 IMAGE" \
-  com.sonatype.license="Apache License, Version 2.0" \
   com.sonatype.name="Nexus IQ Server base image" \
   io.k8s.description="Nexus IQ Server is a policy engine powered by precise intelligence on open source components. \
     It provides a number of tools to improve component usage in your software supply chain, allowing you to \
-    automate your processes and achieve accelerated speed to delivery while also increasing product quality"
-    com.sonatype.license="Apache License, Version 2.0" \
+    automate your processes and achieve accelerated speed to delivery while also increasing product quality" \
```
kellyrob99, line 40: is it possible to extract this to an ENV variable to prevent repetition?

Dockerfile.rhel
```diff
@@ -12,47 +12,45 @@
...
+ARG IQ_SERVER_VERSION=1.85.0-01
+ARG IQ_SERVER_SHA256=da8c232db606c4826d6de160eaa6ee137beaa48b54a98ad4034faa14d692060e
 
 # Mandatory parameters. Docker needs to know volume mount point and location of startup script.
 ENV SONATYPE_WORK="/sonatype-work" \
     IQ_HOME="/opt/sonatype/nexus-iq-server/" \
     DOCKER_TYPE="rh-docker"
 
-ARG IQ_SERVER_COOKBOOK_VERSION="release-0.4.20200210-175005.e554f17"
+ARG IQ_SERVER_COOKBOOK_VERSION="release-0.4.20200219-224442.056dc0d"
```
kellyrob99, line 50: do we support being able to override these build args in automation or are they just for test purposes?

fernau (approved):
+1

Dockerfile.rhel
```diff
@@ -12,47 +12,45 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/rhel7/rhel
+FROM registry.access.redhat.com/ubi8/ubi
```
jflinchbaugh, line 15: this makes me wonder in my own PR for docker-nexus3, if I needed to retain the rhel images, or if I really only need ubi.  I kept ubi separate.

Dockerfile.rhel
```diff
@@ -12,47 +12,45 @@
...
 
 MAINTAINER Sonatype <cloud-ops@sonatype.com>
 
 LABEL name="Nexus IQ Server image" \
   vendor=Sonatype \
-  version="1.84.0-01" \
-  release="1.84.0" \
+  version="1.85.0-01" \
+  release="1.85.0-01" \
```
jflinchbaugh: should "release" remain the shorter version (1.85) lacking the build number (-01)? I kept '-01' off the release for nexus3, and it still built fine in red hat's service.

Dockerfile.rhel
jflinchbaugh: since there *is* a rhel platform we used to use, it might be clearer to rename this file Dockerfile.rh.ubi or something.

adrianpowell: > this makes me wonder in my own PR for docker-nexus3, if I needed to retain the rhel images, or if I really only need ubi. I kept ubi separate.

I'm not sure of the value of the rhel7 image. I don't think anyone outside of the RHCC would use it, so I'm going to skip it until there's a demonstrated need, otherwise we might be stuck maintaining it. (Plus I couldn't get it running)

Dockerfile.rhel
adrianpowell: I'm not sure, but I doubt it. I don't actually know the full motivation for using ARG but I tried changing the cookbook url (also an arg) as a part of the RH build steps and found it challenging and eventually gave up so I have found it of little actual utility so far.

Dockerfile.rhel
adrianpowell: > seems like this should be used for the `version` and `release` values set above?

Great suggestion! Done.

Dockerfile.rhel
adrianpowell: > should "release" remain the shorter version (1.85) lacking the build number (-01).

👍  Fixed

Dockerfile.rhel
adrianpowell: > since there _is_ a rhel platform we used to use, it might be clearer to rename this file Dockerfile.rh.ubi or something.

I don't see so much value in the `.ubi` since both images are from `ubi`, but I like `.rh` (dropping the `el` which might be confusing). Is that okay with you?

PR #40, opened 2020-04-27
use support email instead of defunct cloud-ops

jflinchbaugh:
cloud-ops@s.c. email bounces, so use support instead.


fernau (approved):
LGTM

PR #41, opened 2020-05-01
INT-2562 trigger red hat build

jflinchbaugh:
When the IQ docker image is built, kick off the red hat certified build as well, just like NXRM 3.

JIRA: https://issues.sonatype.org/browse/INT-2562

Jenkins: https://jenkins.ci.sonatype.dev/job/integrations/job/cloud/job/docker-nexus-iq-server-feature/job/INT-2562-trigger-rh-build/3/
kicked off and built this red hat image: https://access.redhat.com/containers/#/registry.connect.redhat.com/sonatype/nexus-iq-server

### DIFF
--- a/ci/TriggerRedHatBuild.groovy
+++ b/ci/TriggerRedHatBuild.groovy
@@ -1,0 +1,192 @@
+/*
+ * Copyright (c) 2020-present Sonatype, Inc. All rights reserved.
+ * Includes the third-party code listed at http://links.sonatype.com/products/clm/attributions.
+ * "Sonatype" is a trademark of Sonatype, Inc.
+ */
+
+/**
+ * This script triggers the build service for a certified docker image at Red Hat.
+ * It's meant to be used by Jenkins via the Jenkinsfile.
+ */
+@Grab('io.github.http-builder-ng:http-builder-ng-core:1.0.4')
+
+import groovyx.net.http.HttpBuilder
+import groovyx.net.http.HttpException
+
+if (args.size() < 3) {
+  System.err.println('Usage: groovy TriggerRedhatBuild.groovy <version> <projectId> <apiKey>')
+  System.exit(1)
+}
+
+new BuildClient(*args).run()
+
+class BuildClient {
+  private static final Integer TIMEOUT_MINUTES = 20
+
+  private final String version
+  private final String projectId
+
+  private final HttpBuilder builder
+
+  BuildClient(String version, String projectId, String apiKey) {
+    this.version = version
+    this.projectId = projectId
+
+    builder = HttpBuilder.configure {
+      request.uri = 'https://connect.redhat.com'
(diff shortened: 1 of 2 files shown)


ci/TriggerRedHatBuild.groovy
```diff
@@ -0,0 +1,192 @@
+/*
```
jflinchbaugh, line 1: Copied from docker-nexus3 repo. When I find a 3rd use, then we can roll this into a shared pipeline or something.

ci/TriggerRedHatBuild.groovy
```diff
@@ -0,0 +1,192 @@
...
+  private String getNextTag(String version) {
+    final tags = getTags()*.name.collectMany {
+      it.split(', ').collect()
+    }
+
+    final currentIndex = tags.findAll {
+      it.startsWith(version)
+    }.collect {
+      it.replaceAll(/${version}-(\d+)-?.*/, '$1') as Integer
+    }.sort().reverse()[0]
```
kellyrob99, line 119: minor but if we do end up reusing this for future projects, this will throw an exception for anything that has not previously been tagged

ci/TriggerRedHatBuild.groovy
```diff
@@ -0,0 +1,192 @@
...
+  /**
+  * Poll for the completed (built and scanned) build at Red Hat build service.
+  * @param nextTag the full version tag assigned to the new build
+  * @return the map from json with info about the completed build
+  */
+  private Map getCompletedBuild(String nextTag) {
+    final endTime = calcCutoffTime(System.currentTimeMillis(), TIMEOUT_MINUTES)
+
+    while (System.currentTimeMillis() < endTime) {
+      println 'Waiting for build to finish.'
```
kellyrob99: might want to include the sleep time in this message to make the interval clear to anyone trailing logs?

ci/TriggerRedHatBuild.groovy
jflinchbaugh: I thought I had accounted for that: 
```
groovy:000> [].sort().reverse()[0]
===> null
```
And then the null gets turned into a 0 down on line 120. Did you see something else of concern?

kellyrob99: hmmm, I guess Groovy got better at this - I expected an IndexOutOfBoundsException given we're referencing the first element of an empty list
My bad!

jflinchbaugh: I always have to test it and remind myself what works and doesn't. I checked it in older Groovy 2.5.6 as well.

ci/TriggerRedHatBuild.groovy
jflinchbaugh: nice idea. implemented as a constant and message.

kellyrob99 (approved):

adrianpowell (approved):

PR #42, opened 2020-05-20
CLM-15817 ; Add in scanning

timlevett:
This PR adds in an IQ evaluation into the docker pipeline.

+ `build` stage for non-master builds
+ `stage-release` stage for master but not pushing
+ `release` stage for master builds (since master builds ship to docker hub)

This also changes `skip_push` to be `push_image` and is false by default which is a change in behavior

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,6 +26,7 @@ node('ubuntu-zion') {
       credentialsId = 'integrations-github-api',
       imageName = 'sonatype/nexus-iq-server',
       archiveName = 'docker-nexus-iq-server',
+      iqApplicationId = 'docker-nexus-iq-server',
       dockerHubRepository = 'nexus-iq-server',
       tarName = 'docker-nexus-iq-server.tar'
   GitHub gitHub
@@ -108,13 +109,10 @@ node('ubuntu-zion') {
       OsTools.runSafe(this, "docker save ${imageName} -o ${env.WORKSPACE}/${tarName}")
       
       //decide which stage we are creating
-      def theStage = 'build'
-      if(branch == 'master') {
-        theStage = 'release'
-      }
+      def theStage = branch == 'master' ? 'release' : 'build'
 
       //run the evaluation
-      nexusPolicyEvaluation iqStage: "${theStage}", iqApplication: "${archiveName}",
+      nexusPolicyEvaluation iqStage: theStage, iqApplication: iqApplicationId,
         iqScanPatterns: [[scanPattern: '*.tar']],
         failBuildOnNetworkError: true
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -106,7 +106,7 @@ node('ubuntu-zion') {
       OsTools.runSafe(this, "docker save ${imageName} -o ${env.WORKSPACE}/${tarName}")
 
       nexusPolicyEvaluation
-        failBuildOnNextworkError: true,
+        failBuildOnNetworkError: true,
         iqApplication: 'docker-nexus-iq-server',
         iqScanPatterns: [[scanPattern: '*.tar']],
         iqStage: 'develop'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,8 @@ node('ubuntu-zion') {
       credentialsId = 'integrations-github-api',
       imageName = 'sonatype/nexus-iq-server',
       archiveName = 'docker-nexus-iq-server',
-      dockerHubRepository = 'nexus-iq-server'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -101,6 +102,23 @@ node('ubuntu-zion') {
...
+    stage('Evaluate') {
+      
+      //Create tar of our image
+      OsTools.runSafe(this, "docker save ${imageName} -o ${env.WORKSPACE}/${tarName}")
+      
+      //decide which stage we are creating
+      def theStage = 'build'
+      if(branch == 'master') {
+        theStage = 'release'
+      }
```
tneer: It's just a style thing so feel free to disregard, but I feel like we usually use the ternary operator for this sort of thing.

```groovy
def theStage = branch == 'master' ? 'release' : 'build'
```

Jenkinsfile
```diff
@@ -101,6 +102,23 @@ node('ubuntu-zion') {
...
+
+      //run the evaluation
+      nexusPolicyEvaluation iqStage: "${theStage}", iqApplication: "${archiveName}",
```
tneer: It shouldn't be necessary to use Groovy string interpolation here.  You can access the variables directly.

```groovy
nexusPolicyEvaluation iqStage: theStage, iqApplication: archiveName,
```

bentmann: I don't object to using "docker-nexus-iq-server" as the IQ app id but pulling that value from a variable named "archiveName" which is otherwise used well for the tarball seems dangerous: Somebody later tweaking that value above at the beginning of the script might be surprised to find this affected policy evaluation as well... Hence I suggest to introduce a dedicated variable `iqApplicationId` or similar which can be set close to https://github.com/sonatype/docker-nexus-iq-server/blob/7e18ade46aa03bf7d9db34ea664e27fb400b553d/Jenkinsfile#L28 where all the other variables are initialized.

Jenkinsfile
```diff
@@ -101,6 +102,23 @@ node('ubuntu-zion') {
         gitHub.statusUpdate commitId, 'success', 'test', 'Tests succeeded'
       }
     }
+    stage('Evaluate') {
+      
+      //Create tar of our image
+      OsTools.runSafe(this, "docker save ${imageName} -o ${env.WORKSPACE}/${tarName}")
+      
+      //decide which stage we are creating
```
bentmann, line 111: Something we should think about: Do we really want to fail the docker image build if something in the base image (as opposed to the app itself) violates policies? At least initially, when quite a bit of catchup/cleanup might be in order, this seems to risk not being able to deliver a docker image for some time.

targeter21: Anecdotally, RedHat (as I've been told) treats base image violations very seriously due to some of their customers who use it and historically pushes UBI fixes very quickly. In my opinion, I think we should not fail the build until we see how bad the violation occurrence is.

timlevett: The alternative would be to ship a known vulnerable package, I think I'm okay with a delayed image. Should we ship something we know has security concerns? If you check out the latest scan, it doesn't scan the IQ jar or the CLI jar because they are uber jars.

Jenkinsfile
timlevett: Fixed in b6aaba7

Jenkinsfile
timlevett: Fixed in b6aaba7

timlevett: Added new variable in b6aaba7

Jenkinsfile
bentmann: >  I think I'm okay with a delayed image

Please run this by and approved by the larger leadership team. In the past, folks had a strong desire to not call a release done until the docker images were also updated. I don't want us tackling the next release and then folks coming out of the bushes asking why the docker images aren't available for days, or even weeks, until everything wrong with the base image is resolved.

>  Should we ship something we know has security concerns?

This might be a good use of IQ's grandfathering feature, take the status quo as baseline, improve upon on it as resources allow, instead of stopping everything.

bentmann (approved):
+1, no general objection to the changes here once we have that conceptual issue resolved how to deal with the current state of affairs of the base image.

mealingr (approved):
+1

Jenkinsfile
timlevett: I discussed the idea of using Grandfathering with Mike and Brian and they both agreed that that is a good approach to this scenario. Therefore, I shall do that.

timlevett: Also discussed blocking the build, we all agreed we should block the push if there is an issue that is found during the release process. We hope that continuous monitoring will prevent said scenario.

PR #47, opened 2020-09-15
Don't mention setting the proxy via config since it's now set via REST/UI - CLM-16695

mealingr:
Bug: https://issues.sonatype.org/browse/CLM-16695


vladt (approved):
+1

anna-d (approved):
+1

PR #48, opened 2020-09-28
Remove chef from the build - CLM-16732

mealingr:
Story: https://issues.sonatype.org/browse/CLM-16732
Build: https://jenkins.ci.sonatype.dev/job/integrations/job/cloud/job/docker-nexus-iq-server-feature/job/CLM-16732_remove_chef/lastBuild
Similar to: https://github.com/sonatype/docker-nexus-iq-server/pull/30


Dockerfile.rh
```diff
@@ -12,14 +12,10 @@
...
 
 MAINTAINER Sonatype <support@sonatype.com>
 
-# Optional parameters.
-ARG IQ_SERVER_VERSION=1.99.0-01
-ARG IQ_SERVER_SHA256=b8e13545503bf4b8a93a7918fc6a56e330a620c1533f91cd9ffcb4bbaaee23c5
-
 LABEL name="Nexus IQ Server image" \
   vendor=Sonatype \
   version="1.99.0-01" \
```
tneer: Just wondering if you can use an ARG value in the LABEL so you don't have to update the version in multiple places.

mealingr: > Just wondering if you can use an ARG value in the LABEL so you don't have to update the version in multiple places.

Yeah you can https://github.com/sonatype/docker-nexus-iq-server/pull/48/commits/5e1e59d9aee5554ea119c2c9260edae516b014a7, I don't think we can change the value to get the release version though, seem to only be allowed to use a constant or an argument/environment variable

anna-d (approved):
+1

Dockerfile
```diff
@@ -12,48 +12,78 @@
...
-    && yum clean all
+# Create folders
+RUN mkdir -p ${TEMP} \
+&& mkdir -m 0755 -p ${IQ_HOME} \
+&& mkdir -m 0755 -p ${SONATYPE_WORK} \
+&& mkdir -m 0755 -p ${CONFIG_HOME} \
+&& mkdir -m 0755 -p ${LOGS_HOME}
 
+# Copy config.yml and set sonatypeWork to the correct value
+COPY config.yml ${TEMP}
```
kakumara, line 47: To my knowledge: From where is this config.yml placed here (in to the running directory)?

Dockerfile.rh
```diff
@@ -40,38 +47,71 @@ LABEL name="Nexus IQ Server image" \
...
+&& rm -rf ${TEMP}
+
+# Add group and user
+RUN groupadd nexus \
+&& adduser -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false -r nexus
+
+# Red Hat Certified Container commands
+COPY rh-docker /
+RUN usermod -a -G root nexus \
+&& chmod -R 0755 /licenses \
```
kakumara, line 89: Q: why do we need this?

Dockerfile
kakumara: > To my knowledge: From where is this config.yml placed here (in to the running directory)?

Ah, ignore ☝️ . I can see you have added it below in the root.

mealingr: > To my knowledge: From where is this config.yml placed here (in to the running directory)?

The `config.yml` is being added in this PR, in the same location as the `Dockerfile` so as long as the running directory is where the `Dockerfile` is this should be correct

Edit: apologies responded before i refreshed

Dockerfile.rh
mealingr: > Q: why do we need this?

This particular line mimics what was previously being done by chef
https://github.com/sonatype/chef-nexus-iq-server/blob/601fa0417bb6a9e6fe50bfca5a29ff3453ec56d6/recipes/rh-docker.rb#L24-L29

config.yml
```diff
@@ -0,0 +1,37 @@
+sonatypeWork: "/sonatype-work"
```
kakumara, line 1: With this, do we now need to maintain this to keep it in sync with insight-brain repo?

mealingr: > With this, do we now need to maintain this to keep it in sync with insight-brain repo?

Somewhat, it doesn't need to be an exact duplicate of `config.yml` in insight-brain, but if we update the docker image to use a new brain version, then we need to make sure the `config.yml` still makes sense. This is similar to the situation with chef (the only difference is chef was programmatically creating its `config.yml` but whatever it created was equally at risk of being outdated)

Dockerfile.rh
kakumara: > > Q: why do we need this?
> 
> This particular line mimics what was previously being done by chef
> https://github.com/sonatype/chef-nexus-iq-server/blob/601fa0417bb6a9e6fe50bfca5a29ff3453ec56d6/recipes/rh-docker.rb#L24-L29

I'm assuming this directory is for storing the license? Is it correct to say with the latest version of IQ we now store the license in database?

config.yml
kakumara: > > With this, do we now need to maintain this to keep it in sync with insight-brain repo?
> 
> Somewhat, it doesn't need to be an exact duplicate of `config.yml` in insight-brain, but if we update the docker image to use a new brain version, then we need to make sure the `config.yml` still makes sense. This is similar to the situation with chef (the only difference is chef was programmatically creating its `config.yml` but whatever it created was equally at risk of being outdated)

(Just throwing out a wild idea, feel free to ignore). What if we we keep this in insight-brain just like other config.yml versions (say `config-docker.yml` or something) so that we can maintain it along with PRs if required. And if we package it in the root of the IQ distribution, we should be able to grab it after the extracting the downloaded distribution above.

Dockerfile.rh
mealingr: > I'm assuming this directory is for storing the license? Is it correct to say with the latest version of IQ we now store the license in database?

Not for the IQ license but for this license https://github.com/sonatype/chef-nexus-iq-server/blob/601fa0417bb6a9e6fe50bfca5a29ff3453ec56d6/files/rh-docker/licenses/LICENSE which is unique to the Red Hat build

kakumara: > > I'm assuming this directory is for storing the license? Is it correct to say with the latest version of IQ we now store the license in database?
> 
> Not for the IQ license but for this license https://github.com/sonatype/chef-nexus-iq-server/blob/601fa0417bb6a9e6fe50bfca5a29ff3453ec56d6/files/rh-docker/licenses/LICENSE which is unique to the Red Hat build

👍

config.yml
mealingr: > (Just throwing out a wild idea, feel free to ignore). What if we we keep this in insight-brain just like other config.yml versions (say `config-docker.yml` or something) so that we can maintain it along with PRs if required. And if we package it in the root of the IQ distribution, we should be able to grab it after the extracting the downloaded distribution above.

I can see the advantage (easier for us to maintain) but there's also the disadvantage of some customers may not care about docker, it's unfortunate the files are different. Personally i can see it working either way. On the plus side we generally keep backwards compatibility with older `config.yml` files so the chance of a given `config.yml` file breaking here should be low

kakumara (approved):
+1, LGTM in general

PR #51, opened 2020-11-03
update latest packages

jflinchbaugh:
while removing chef, a `microdnf update` was newly introduced, but that caused some conflicts between versions of rpm-lib, and blocked the build. we eliminated the update to proceed there, but having latest packages could be nice, so this brings back an update that wortks.

build: https://jenkins.ci.sonatype.dev/job/integrations/job/cloud/job/docker-nexus-iq-server-feature/job/update-latest-packages/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,8 @@ LABEL vendor=Sonatype \
 USER root
 
 # For testing
-RUN microdnf install procps
+RUN dnf update --allowerasing -y \
+&& dnf install -y procps
 
 # Create folders
 RUN mkdir -p ${TEMP} \
--- a/Dockerfile.rh
+++ b/Dockerfile.rh
@@ -50,7 +50,8 @@ LABEL name="Nexus IQ Server image" \
 USER root
 
 # For testing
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -33,7 +33,8 @@ LABEL vendor=Sonatype \
 USER root
 
 # For testing
-RUN microdnf install procps
+RUN dnf update --allowerasing -y \
+&& dnf install -y procps
```
jflinchbaugh, line 37: this works, but there could be better approaches.

mealingr: > this works, but there could be better approaches.

I see `--allowerasing` seems like the essential part, and `microdnf` doesn't seem to support it whereas regular full blown `dnf` does

mealingr (approved):
+1

eduard-tita (approved):
:+1: LGTM

jflinchbaugh:
I thought maybe this PR introduced a pile of policy violations, but it appears they were already in master from a previous change. https://issues.sonatype.org/browse/CLM-17310

PR #53, opened 2020-11-05
Escape JAVA_OPTS so it is not evaluated during build time

mrook:


hboutemy:
it seems this is a little typo introduced in #48 
@mealingr can you have a look, please?
I didn't personally test the fix, but can confirm the initial bug: `start.sh` script does not contain the expected `${JAVA_OPTS}` but its empty evaluation from image build time

mealingr:
> it seems this is a little typo introduced in #48
> @mealingr can you have a look, please?
> I didn't personally test the fix, but can confirm the initial bug: `start.sh` script does not contain the expected `${JAVA_OPTS}` but its empty evaluation from image build time

Took a look and you're right, also tested this and it does seem to fix it, thanks @mrook for spotting this

mealingr (approved):
+1

jamengual:
interesting.

we use a similar method to this: https://stackoverflow.com/questions/53785577/passing-java-opts-to-spring-boot-application-through-docker-compose

you do not need to echo to a file to set the JAVA_OPTS

mrook:
@mealingr happy to help!

hboutemy:
I suppose the same update has to be made into `Dockerfile.rh`, even if I don't know when this Dockerfile is used

> you do not need to echo to a file to set the JAVA_OPTS

in fact, we echo to `start.sh` script, which is the script used as Dockerfile `ENTRYPOINT`

mealingr:
> I suppose the same update has to be made into `Dockerfile.rh`

I think you're right @hboutemy, @mrook I've merged your change for both Dockerfiles in https://github.com/sonatype/docker-nexus-iq-server/commit/af5db6f15195c0e6c6d84f83dcff80ac85edda96 and created a PR https://github.com/sonatype/docker-nexus-iq-server/pull/56 to update the docs with a workaround for 101 and 102, thanks again

mrook:
Cool, thanks!

PR #54, opened 2020-11-05
Specify uid and gid to prevent need to migrate ownership on base image change - CLM-17312

mealingr:
Bug: https://issues.sonatype.org/browse/CLM-17312

Starting IQ 101 via
`docker run -d -p 8070:8070 -p 8071:8071 --name nexus-iq-server sonatype/nexus-iq-server:1.101.0`

And checking the uid of the `nexus` user and the gid of the `nexus` group gives
```
id nexus
uid=997(nexus) gid=1000(nexus) groups=1000(nexus)
```
This PR specifically sets them to these values, which should prevent needing to take ownership of existing volumes when migrating in the future if the base image changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,8 +62,8 @@ RUN cd ${TEMP} \
 && rm -rf ${TEMP}
 
 # Add group and user
-RUN groupadd nexus \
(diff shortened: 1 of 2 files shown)


shaikhu (approved):
+1

mr-zepol (approved):
+1

PR #55, opened 2020-11-10
Add Jenkinsfile to push README

mealingr:
Context: https://sonatype.slack.com/archives/CCBVDJZM4/p1605011621459900

### DIFF
--- a/Jenkinsfile.updateReadme
+++ b/Jenkinsfile.updateReadme
@@ -9,6 +9,7 @@ import com.sonatype.jenkins.pipeline.OsTools
 
 node('ubuntu-zion') {
   def branch,
+      readme,
       organization = 'sonatype',
       dockerHubRepository = 'nexus-iq-server'
   try {
@@ -16,6 +17,10 @@ node('ubuntu-zion') {
       deleteDir()
       def checkoutDetails = checkout scm
       branch = checkoutDetails.GIT_BRANCH == 'origin/master' ? 'master' : checkoutDetails.GIT_BRANCH
+      readme = readFile file: 'README.md', encoding: 'UTF-8'
+      readme = readme.replaceAll("(?s)<!--.*?-->", "")
+      readme = readme.replace("\"", "\\\"")
+      readme = readme.replace("\n", "\\n")
     }
     if (branch == 'master') {
       stage('Push README.md') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.updateReadme
```diff
@@ -0,0 +1,56 @@
...
+              url: "https://hub.docker.com/v2/repositories/${organization}/${dockerHubRepository}/"
+        }
+      }
+    }
+    else {
+      stage('Check README.md') {
+        def readme = readFile file: 'README.md', encoding: 'UTF-8'
+            readme = readme.replaceAll("(?s)<!--.*?-->", "")
+            readme = readme.replace("\"", "\\\"")
+            readme = readme.replace("\n", "\\n")
```
jflinchbaugh: Could the readme be processed before checking master branch, since you want it unconditionally? at least a private method?

Jenkinsfile.updateReadme
```diff
@@ -0,0 +1,56 @@
...
+            echo readme
+      }
+    }
+  } finally {
+    OsTools.runSafe(this, "docker logout")
+    OsTools.runSafe(this, 'git clean -f && git reset --hard origin/master')
```
jflinchbaugh: Q: I wonder if anything is really changed by this little job, so is this necessary?

Jenkinsfile.updateReadme
mealingr: > Could the readme be processed before checking master branch, since you want it unconditionally? at least a private method?

That makes more sense, done in https://github.com/sonatype/docker-nexus-iq-server/pull/55/commits/8122b877b52f21522169ec9607eb1730e6e04d77

Jenkinsfile.updateReadme
mealingr: > Q: I wonder if anything is really changed by this little job, so is this necessary?

I don't think it's needed since there's no changes, removed in https://github.com/sonatype/docker-nexus-iq-server/pull/55/commits/8122b877b52f21522169ec9607eb1730e6e04d77

jflinchbaugh (approved):
+1 looks like a convenient capability. I'll have to remember to go this route instead of all the switch parameters in one file that I've used previously. 👍

collinpeters (approved):
+1

FWIW there was chatter recently on updating the readme on https://hub.docker.com/r/sonatype/nexus-platform-cli (which is a base image for two other integrations) but we stopped because of the hurdles. This could become a good *generic* solution so perhaps in the future we can make this a generic Jenkins job where you provide the GH repo and Hub namespace and it pushes the readme. Just mentioning for future ideas.

PR #57, opened 2020-11-26
Reduce image size

hboutemy:
looking at compressed image size on DockerHub https://hub.docker.com/r/sonatype/nexus-iq-server/tags?page=1&ordering=last_updated
- release 100: 273.32 MB
- release 101: 383.21 MB
- release 102: 523.75 MB

Found 2 causes:
1. nexus-iq-server.jar (~130 MB) is stored twice because chown is done in a separate RUN
2. while installing procps, there is also an update of packages, which brings in a new release of OpenJDK

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,12 +36,14 @@ USER root
 RUN dnf update --allowerasing -y \
 && dnf install -y procps
 
-# Create folders
+# Create folders, add group and user
 RUN mkdir -p ${TEMP} \
 && mkdir -m 0755 -p ${IQ_HOME} \
 && mkdir -m 0755 -p ${SONATYPE_WORK} \
 && mkdir -m 0755 -p ${CONFIG_HOME} \
-&& mkdir -m 0755 -p ${LOGS_HOME}
+&& mkdir -m 0755 -p ${LOGS_HOME} \
+&& groupadd -g 1000 nexus \
+&& adduser -u 997 -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false -r nexus
 
 # Copy config.yml and set sonatypeWork to the correct value
 COPY config.yml ${TEMP}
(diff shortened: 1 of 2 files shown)


mealingr (approved):
+1 (tested these changes on another branch and the build passed https://jenkins.ci.sonatype.dev/job/integrations/job/cloud/job/docker-nexus-iq-server-feature/job/reduce_image_size/lastBuild)

bentmann:
> there is also an update of packages

That update was very recently added in https://github.com/sonatype/docker-nexus-iq-server/pull/51, the authors should get together to clarify the intent/ramfications.

From the outside, not using latest versions seems to increase the risk of distributing vulnerable versions.

mealingr:
> From the outside, not using latest versions seems to increase the risk of distributing vulnerable versions.

Fair point, it might be better to add the
```
RUN dnf update --allowerasing -y \
&& dnf install -y procps
```
lines back in for now and get the other changes through, can always revisit the package updates later

hboutemy:
I don't know what is the policy of RH base images regarding updates: if base images are not updated, yes, every consuming image will have to update OS packages provided by base (with implied duplicate)

notice that this question has been left unnoticed until adding `procps` package (for automated testing purpose): I suppose it was just that when adding a package, you usually update your distribution before
In 1.102.0 release, this updated the JDK, which is really a big package.
But in the 1.103.0 release, it seems base image already contained the update, then the duplicate size cost was really transient...

but for sure, the question of updating or not OS package can be worked on in a separate stream: chown of nexus-iq-server.jar is a useful improvement

hboutemy:
@mealingr do you want me to drop the second commit (remove dnf update) from the current PR?

mealingr:
> @mealingr do you want me to drop the second commit (remove dnf update) from the current PR?

@hboutemy I would vote to drop it for now to get the other changes merged faster

hboutemy:
@mealingr no problem, I did 2 commits because I knew that it was 2 completely separate aspects

I just removed the second commit: you'll decide later if you want to remove the `dnf update` or not

Dockerfile
```diff
@@ -60,14 +62,8 @@ RUN cd ${TEMP} \
 && tar -xvf nexus-iq-server-${IQ_SERVER_VERSION}-bundle.tar.gz \
 && mv nexus-iq-server-${IQ_SERVER_VERSION}.jar ${IQ_HOME} \
 && cd ${IQ_HOME} \
-&& rm -rf ${TEMP}
-
-# Add group and user
-RUN groupadd -g 1000 nexus \
-&& adduser -u 997 -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false -r nexus
-
-# Change owner to nexus user
```
bentmann, line 69: While the changes here generally work, I'd like to propose a slightly different formatting that allows to preserve the logical blocks along with their commentary (cf. https://docs.docker.com/engine/reference/builder/#format):
https://github.com/sonatype/docker-nexus-iq-server/commit/189523172da892e1eec0e3e3673d712f0973cc45

hboutemy: oh, nice: if comments can be kept like this, I'm full +1

bentmann: Would you like to incorporate that technicality into your PR and we merge that then or shall I rather just run with my commit? I don't want to make it more complicated for you but ensure you don't feel your thunder has been stolen.

hboutemy: no problem, just merge your commit, it's simpler: I know you well, you didn't steal anything but added value :)

PR #58, opened 2020-12-07
use jenkins-shared explicitly

jflinchbaugh:
jenkins shared is no longer magically added, so explicitly include it. also switch to private pipeline.


tneer (approved):
+1 LGTM

PR #59, opened 2020-12-08
Fix location of audit.log

bentmann:
https://issues.sonatype.org/browse/CLM-17366

By the default configuration, the audit log is created at `./log/audit.log` which for the docker image translates to `/opt/sonatype/nexus-iq-server/log/audit.log`. However, the log files are expected to reside in `/var/log/nexus-iq-server/`, so these changes explicitly configure that location for the audit log as well.


kakumara (approved):
+1, LGTM

anna-d (approved):
+1

PR #60, opened 2020-12-08
Health check

bentmann:
https://issues.sonatype.org/browse/CLM-17495

As background:
- https://docs.docker.com/engine/reference/builder/#healthcheck
- https://curl.se/docs/manpage.html#-f

Unlike suggested in JIRA, this here pings the main port which is a more relevant indicator of the app's health. `--fail` tells curl to signal HTTP errors via its exit code which `|| exit 1` converts from 22 to 1 to comply with the Docker spec.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,9 @@ VOLUME ${LOGS_HOME}
 EXPOSE 8070
 EXPOSE 8071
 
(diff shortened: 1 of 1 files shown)


targeter21 (approved):
+1

mealingr (approved):
+1

PR #62, opened 2020-12-09
Synchronize the LABEL data

bentmann:
I'm looking for one of two things: Either somebody from the original set of authors educates me on the need for subtle differences between the LABEL instructions or we can quickly agree this change is a good thing.

From https://docs.docker.com/engine/reference/builder/#label and what I see over at https://github.com/sonatype/docker-nexus3, I currently suspect those differences are unintentional.

### DIFF
--- a/Dockerfile.rh
+++ b/Dockerfile.rh
@@ -23,9 +23,8 @@ ARG SONATYPE_WORK="/sonatype-work"
 ARG CONFIG_HOME="/etc/nexus-iq-server"
 ARG LOGS_HOME="/var/log/nexus-iq-server"
 
-MAINTAINER Sonatype <support@sonatype.com>
-
 LABEL name="Nexus IQ Server image" \
+  maintainer="Sonatype <support@sonatype.com>" \
   vendor=Sonatype \
   version="${IQ_SERVER_VERSION}" \
   release="1.103.0" \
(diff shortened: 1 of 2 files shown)


Dockerfile.rh
```diff
@@ -23,9 +23,8 @@ ARG SONATYPE_WORK="/sonatype-work"
 ARG CONFIG_HOME="/etc/nexus-iq-server"
 ARG LOGS_HOME="/var/log/nexus-iq-server"
 
-MAINTAINER Sonatype <support@sonatype.com>
```
bentmann, line 26: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated, so moved into LABEL.

Dockerfile.rh
```diff
@@ -37,7 +36,6 @@ LABEL name="Nexus IQ Server image" \
   com.sonatype.license="Apache License, Version 2.0" \
   com.sonatype.name="Nexus IQ Server image" \
   run="docker run -d -p 8070:8070 -p 8071:8071 IMAGE" \
-  com.sonatype.name="Nexus IQ Server base image" \
```
bentmann, line 40: Duplicates key from two lines above.

eduard-tita (approved):
+1 LGTM

PR #64, opened 2021-01-22
docker push --all-tags

jflinchbaugh:
Docker seems to have recently changed to only push most recent tag by default.
This applies the same wisdom found in https://github.com/sonatype/docker-nexus3/pull/127/files

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -147,7 +147,7 @@ node('ubuntu-zion') {
           OsTools.runSafe(this, """
             docker login --username ${env.DOCKERHUB_API_USERNAME} --password ${env.DOCKERHUB_API_PASSWORD}
           """)
-          OsTools.runSafe(this, "docker push ${organization}/${dockerHubRepository}")
+          OsTools.runSafe(this, "docker push --all-tags ${organization}/${dockerHubRepository}")
 
           response = OsTools.runSafe(this, """
             curl -X POST https://hub.docker.com/v2/users/login/ \


Jenkinsfile
```diff
@@ -147,7 +147,7 @@ node('ubuntu-zion') {
           OsTools.runSafe(this, """
             docker login --username ${env.DOCKERHUB_API_USERNAME} --password ${env.DOCKERHUB_API_PASSWORD}
           """)
-          OsTools.runSafe(this, "docker push ${organization}/${dockerHubRepository}")
+          OsTools.runSafe(this, "docker push --all-tags ${organization}/${dockerHubRepository}")
```
fernau, line 150: thanks for figuring this out!  it already works for docker-nexus3 👍

fernau (approved):
LGTM

Jenkinsfile
fernau: is there a docker shared pipeline somewhere?

jflinchbaugh: you mean a place for these common steps of tagging and pushing to live? I've not seen it.

fernau: i'm wondering where else the same problem is still lurking around... could do a search across github.. not saying that's the best way to start this

jflinchbaugh: i intend to visit docker-nexus-iq-cli for good measure, but that one seems to have somehow done the right thing in this last release.

PR #67, opened 2021-04-19
Added optional build arguments CLM-17772

anna-d:
Jira: https://issues.sonatype.org/browse/CLM-17772

Two optional build arguments have been added, `UID` and `GID`. They allow the user to override the default user and/or group that are used when running the image.

These arguments can be used as follows:
```
docker build --build-arg UID=1000 --build-arg GID=1000 -t nexus-iq-server . 
```

Using these arguments when building the image, will prevent the "Permission denied" error  when running the container (as described in the description of [CLM-17772](https://issues.sonatype.org/browse/CLM-17772)).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,8 @@ ARG IQ_HOME="/opt/sonatype/nexus-iq-server"
 ARG SONATYPE_WORK="/sonatype-work"
 ARG CONFIG_HOME="/etc/nexus-iq-server"
 ARG LOGS_HOME="/var/log/nexus-iq-server"
+ARG GID=1000
+ARG UID=997
 
 LABEL name="Nexus IQ Server image" \
   maintainer="Sonatype <support@sonatype.com>" \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -76,8 +78,8 @@ RUN cd ${TEMP} \
 && rm -rf ${TEMP} \
 \
 # Add group and user
-&& groupadd -g 1000 nexus \
-&& adduser -u 997 -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false -r nexus \
+&& groupadd -g ${GID} nexus \
+&& adduser -u ${UID} -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false -r nexus \
```
mealingr, line 82: There's also `Dockerfile.rh` which we should make the same changes in I think

mealingr (approved):
+1

Dockerfile
anna-d: > There's also `Dockerfile.rh` which we should make the same changes in I think

Updated in 7598e14941069b198c010f49a978cf4723e43bc8.

mr-zepol (approved):
+1 LGTM

PR #68, opened 2021-06-01
CLM-18632: modify user and remove -r flag

amiralig:
For the full conversation regarding this you can refer to:

https://sonatype.slack.com/archives/CC2BN7KU6/p1622485542189800

We're trying to mount the drive that contains the lock to remove it (sonatype which is mapped on amazon EFS to be accessed by the IQ container) but we're running into a permission issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ARG SONATYPE_WORK="/sonatype-work"
 ARG CONFIG_HOME="/etc/nexus-iq-server"
 ARG LOGS_HOME="/var/log/nexus-iq-server"
 ARG GID=1000
-ARG UID=997
+ARG UID=1000
 
 LABEL name="Nexus IQ Server image" \
   maintainer="Sonatype <support@sonatype.com>" \
@@ -79,7 +79,7 @@ RUN cd ${TEMP} \
 \
 # Add group and user
 && groupadd -g ${GID} nexus \
-&& adduser -u ${UID} -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false -r nexus \
+&& adduser -u ${UID} -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false nexus \
 \
 # Change owner to nexus user
 && chown -R nexus:nexus ${IQ_HOME} \
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,8 @@
 FROM registry.access.redhat.com/ubi8/openjdk-8:1.3-8
 
 # Build parameters
-ARG IQ_SERVER_VERSION=1.116.0-01
-ARG IQ_SERVER_SHA256=4463213916bd935bb0c9f33fbcf59b6af0d7fc13a362ba0cb2041a1be30c6330
+ARG IQ_SERVER_VERSION=1.117.0-01
+ARG IQ_SERVER_SHA256=370a613c1eec34c7e22cf90550247c94076dbd427a895b4a9e546a903e54b307
 ARG TEMP="/tmp/work"
 ARG IQ_HOME="/opt/sonatype/nexus-iq-server"
 ARG SONATYPE_WORK="/sonatype-work"
@@ -29,7 +29,7 @@ LABEL name="Nexus IQ Server image" \
   maintainer="Sonatype <support@sonatype.com>" \
   vendor=Sonatype \
   version="${IQ_SERVER_VERSION}" \
-  release="1.116.0" \
+  release="1.117.0" \
   url="https://www.sonatype.com" \
   summary="The Nexus IQ Server" \
   description="Nexus IQ Server is a policy engine powered by precise intelligence on open source components. \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,8 +13,6 @@ properties([
     string(defaultValue: '', description: 'New Nexus IQ Version Sha256', name: 'nexus_iq_version_sha'),
 
     booleanParam(defaultValue: false, description: 'Push Docker Image and Tags', name: 'push_image'),
-    booleanParam(defaultValue: false, description: 'Force Red Hat Certified Build for a non-master branch', name: 'force_red_hat_build'),
-    booleanParam(defaultValue: false, description: 'Skip Red Hat Certified Build', name: 'skip_red_hat_build'),
   ])
 ])
 
@@ -182,16 +180,6 @@ node('ubuntu-zion') {
       }
     }
 
-    if ((! params.skip_red_hat_build) && (branch == 'master' || params.force_red_hat_build)) {
-      stage('Trigger Red Hat Certified Image Build') {
-        withCredentials([
-            string(credentialsId: 'docker-nexus-iq-rh-build-project-id', variable: 'PROJECT_ID'),
-            string(credentialsId: 'rh-build-service-api-key', variable: 'API_KEY')]) {
-          final redHatVersion = "${version}-ubi"
-          runGroovy('ci/TriggerRedHatBuild.groovy', [redHatVersion, PROJECT_ID, API_KEY].join(' '))
-        }
-      }
-    }
   } finally {
     OsTools.runSafe(this, "docker logout")
     OsTools.runSafe(this, "docker system prune -a -f")
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ node('ubuntu-zion') {
   def commitId, commitDate, version, imageId, branch, dockerFileLocations
   def organization = 'sonatype',
       gitHubRepository = 'docker-nexus-iq-server',
-      credentialsId = 'integrations-github-api',
+      credentialsId = 'sonaype-ci-github-access-token',
       imageName = 'sonatype/nexus-iq-server',
       archiveName = 'docker-nexus-iq-server',
       iqApplicationId = 'docker-nexus-iq-server',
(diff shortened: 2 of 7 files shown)


amiralig:
@anna-d  do you happen to know what jenkins build will trigger once this is merged and if that can be tested via the snapshot environment?

troby (approved):

Dockerfile
```diff
@@ -23,7 +23,7 @@ ARG SONATYPE_WORK="/sonatype-work"
 ARG CONFIG_HOME="/etc/nexus-iq-server"
 ARG LOGS_HOME="/var/log/nexus-iq-server"
 ARG GID=1000
-ARG UID=997
+ARG UID=1000
```
mealingr, line 26: If we change the UID I think this will cause a re-occurrence of https://issues.sonatype.org/browse/CLM-17312

After reading through the thread https://sonatype.slack.com/archives/CC2BN7KU6/p1622485542189800 I see these as the relevant points
> the Dockerfile tries to create nexus user with UID 997 but that is already in use by systemd-coredump
> system UIDs should not be used as future changes could try to use the same UID

fwiw 997 and 1000 were chosen because that's the defaults that were assigned when starting the image back when these were set
https://github.com/sonatype/docker-nexus-iq-server/pull/54

I agree with not using system UIDs, but I think if we update this we will also have to update
https://help.sonatype.com/display/NXIQM/Upgrading+the+IQ+Server

A workaround could be to build a custom image using the optional build arguments to specify UID and GID, these were added in https://github.com/sonatype/docker-nexus-iq-server/pull/67 and can be used via e.g.
```
docker build --build-arg UID=1000 --build-arg GID=1000 -t nexus-iq-server .
```

amiralig: Hi @mealingr I agree with your points about updating the doc if we decide to go with this approach.

@troby what Richard is suggesting is what I also suggested on our call. That would be a minimal change on our jenkins file to run with the params that Richard is proposing. So now we have to decide which approach to take that has the least impact. 

@mealingr and @troby based on your experience in the past, when we update these UIDs, do a lot of clients get impacted? Is it too high-risk (as in would it potentially create a lot of support tickets for us?)

mealingr: > @mealingr and @troby based on your experience in the past, when we update these UIDs, do a lot of clients get impacted? Is it too high-risk (as in would it potentially create a lot of support tickets for us?)

The original issue where UID and GID changed affected at least 4 customers
https://issues.sonatype.org/browse/CLM-13092
https://sonatype.zendesk.com/agent/tickets/36272
https://sonatype.zendesk.com/agent/tickets/36250
https://sonatype.zendesk.com/agent/tickets/36503
https://sonatype.zendesk.com/agent/tickets/36532

The next time this issue occurred it was first raised as an internal issue
https://issues.sonatype.org/browse/CLM-17312
One customer raised an issue
https://sonatype.zendesk.com/agent/tickets/48153
but it was resolved by then

So I would say there's a chance we might get some customer issues even if the docs are updated. Although if the docs are updated the issues should be easy to resolve.

README.md
```diff
@@ -34,6 +34,18 @@ A Dockerfile for Sonatype Nexus IQ Server, based on [Red Hat Universal Base Imag
...
+Version 1.117.0 of the Docker image changed the UID of the nexus user from 997 to 1000, and changed it from a system account to a user account. This is to minimize the chance of a future collision with other UIDs. If you use this image with persistent data volumes, then for each volume you will need to run the following command:
+```
+docker run -it -u=0 -v [volume name]:[volume container path] sonatype/nexus-iq-server:1.117.0 chown -R nexus:nexus [volume container path]
+```
+e.g.
+```
+docker run -it -u=0 -v sonatype-work:/sonatype-work sonatype/nexus-iq-server:1.117.0 chown -R nexus:nexus /sonatype-work
+```
+This will start up a 1.117.0 IQ server container with root as the user, allowing it to chown the sonatype-work directory and its files to the correct nexus user.
+
```
amiralig, line 48: @mealingr I updated readme.

Dockerfile
amiralig: This is the link for the doc updated: 
https://help.sonatype.com/pages/diffpagesbyversion.action?pageId=330916&selectedPageVersions=25&selectedPageVersions=24

README.md
```diff
@@ -34,6 +34,18 @@ A Dockerfile for Sonatype Nexus IQ Server, based on [Red Hat Universal Base Imag
 
 ## Migration
 
+
+### Upgrading from Version 116 or Earlier to Version 117 or Later
```
mealingr: Since release 117 is already happening
```suggestion
### Upgrading from Version 117 or Earlier to Version 118 or Later
```

README.md
```diff
@@ -34,6 +34,18 @@ A Dockerfile for Sonatype Nexus IQ Server, based on [Red Hat Universal Base Imag
...
+Version 1.117.0 of the Docker image changed the UID of the nexus user from 997 to 1000, and changed it from a system account to a user account. This is to minimize the chance of a future collision with other UIDs. If you use this image with persistent data volumes, then for each volume you will need to run the following command:
```
mealingr:
```suggestion
Version 1.118.0 of the Docker image changed the UID of the nexus user from 997 to 1000, and changed it from a system account to a user account. This is to minimize the chance of a future collision with other UIDs. If you use this image with persistent data volumes, then for each volume you will need to run the following command:
```

README.md
```diff
@@ -34,6 +34,18 @@ A Dockerfile for Sonatype Nexus IQ Server, based on [Red Hat Universal Base Imag
...
+```
+docker run -it -u=0 -v [volume name]:[volume container path] sonatype/nexus-iq-server:1.117.0 chown -R nexus:nexus [volume container path]
```
mealingr:
```suggestion
docker run -it -u=0 -v [volume name]:[volume container path] sonatype/nexus-iq-server:1.118.0 chown -R nexus:nexus [volume container path]
```

README.md
```diff
@@ -34,6 +34,18 @@ A Dockerfile for Sonatype Nexus IQ Server, based on [Red Hat Universal Base Imag
...
+```
+e.g.
+```
+docker run -it -u=0 -v sonatype-work:/sonatype-work sonatype/nexus-iq-server:1.117.0 chown -R nexus:nexus /sonatype-work
```
mealingr:
```suggestion
docker run -it -u=0 -v sonatype-work:/sonatype-work sonatype/nexus-iq-server:1.118.0 chown -R nexus:nexus /sonatype-work
```

README.md
```diff
@@ -34,6 +34,18 @@ A Dockerfile for Sonatype Nexus IQ Server, based on [Red Hat Universal Base Imag
...
+```
+This will start up a 1.117.0 IQ server container with root as the user, allowing it to chown the sonatype-work directory and its files to the correct nexus user.
```
mealingr:
```suggestion
This will start up a 1.118.0 IQ server container with root as the user, allowing it to chown the sonatype-work directory and its files to the correct nexus user.
```

Dockerfile
amiralig: Updated to 118 in the docs as well:
https://help.sonatype.com/pages/diffpagesbyversion.action?pageId=330916&selectedPageVersions=28&selectedPageVersions=27

mealingr (approved):
+1

spec/Dockerfile_spec.rb
```diff
@@ -45,7 +45,7 @@
     end
 
     it 'has a specific id' do
-      expect(subject).to have_uid(997)
+      expect(subject).to have_uid(1000)
```
amiralig, line 48: This one was failing on jenkins

mealingr:
@amiralig I almost forgot I think you need to make the same changes in the `Dockerfile.rh` file as well
https://github.com/sonatype/docker-nexus-iq-server/blob/54f0ee0b2c51afdf865b898c636bd36b3accfd0c/Dockerfile.rh#L26
https://github.com/sonatype/docker-nexus-iq-server/blob/54f0ee0b2c51afdf865b898c636bd36b3accfd0c/Dockerfile.rh#L82

Dockerfile
```diff
@@ -15,21 +15,21 @@
 FROM registry.access.redhat.com/ubi8/openjdk-8:1.3-8
 
 # Build parameters
-ARG IQ_SERVER_VERSION=1.116.0-01
-ARG IQ_SERVER_SHA256=4463213916bd935bb0c9f33fbcf59b6af0d7fc13a362ba0cb2041a1be30c6330
+ARG IQ_SERVER_VERSION=1.117.0-01
```
amiralig, line 18: Do I need to manually set these to higher version?

Jenkinsfile
```diff
@@ -13,16 +13,14 @@ properties([
...
   ])
 ])
 
 node('ubuntu-zion') {
   def commitId, commitDate, version, imageId, branch, dockerFileLocations
   def organization = 'sonatype',
       gitHubRepository = 'docker-nexus-iq-server',
-      credentialsId = 'integrations-github-api',
+      credentialsId = 'sonaype-ci-github-access-token',
       imageName = 'sonatype/nexus-iq-server',
```
amiralig, line 24: these changes came when I rebased

amiralig:
> @amiralig I almost forgot I think you need to make the same changes in the `Dockerfile.rh` file as well
> https://github.com/sonatype/docker-nexus-iq-server/blob/54f0ee0b2c51afdf865b898c636bd36b3accfd0c/Dockerfile.rh#L26
> 
> 
> https://github.com/sonatype/docker-nexus-iq-server/blob/54f0ee0b2c51afdf865b898c636bd36b3accfd0c/Dockerfile.rh#L82


thank you the changes are made

Dockerfile
mealingr: > Do I need to manually set these to higher version?

The version and hashes are automatically updated during a release
https://github.com/sonatype/docker-nexus-iq-server/commit/54f0ee0b2c51afdf865b898c636bd36b3accfd0c

I think something might be wrong with the rebase it shouldn't be showing this as a change since it's already 1.117.0-01 on master (same for the hash below)

amiralig:
I am closing this PR because some of the diffs shown here are coming from before re-base. I will add the same folks to there.

PR #70, opened 2021-06-07
Added Trap for Shutting Down IQ Gracefully

srini-hdp:
Updated startup.sh to trap sigterm for graceful shutdown of IQ

Jira: https://issues.sonatype.org/browse/CLM-18614
Jenkins: https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server-feature/job/CLM-18614-stop-gracefully/lastBuild/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,7 @@ RUN cat ${TEMP}/config.yml | sed -r "s/\s*sonatypeWork\s*:\s*\"?[-0-9a-zA-Z_/\\]
 && chmod 0644 ${CONFIG_HOME}/config.yml
 
 # Create start script
-RUN echo "trap 'kill -TERM \`cut -f1 -d@ $SONATYPE_WORK/lock\`  && sleep 60' SIGTERM" > ${IQ_HOME}/start.sh \
+RUN echo "trap 'kill -TERM \`cut -f1 -d@ ${SONATYPE_WORK}/lock\` && sleep 60' SIGTERM" > ${IQ_HOME}/start.sh \
 && echo "/usr/bin/java \${JAVA_OPTS} -jar nexus-iq-server-${IQ_SERVER_VERSION}.jar server ${CONFIG_HOME}/config.yml 2> ${LOGS_HOME}/stderr.log & " >> ${IQ_HOME}/start.sh \
 && echo "wait" >> ${IQ_HOME}/start.sh \
 && chmod 0755 ${IQ_HOME}/start.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,9 @@ RUN cat ${TEMP}/config.yml | sed -r "s/\s*sonatypeWork\s*:\s*\"?[-0-9a-zA-Z_/\\]
 && chmod 0644 ${CONFIG_HOME}/config.yml
 
 # Create start script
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -64,7 +64,9 @@ RUN cat ${TEMP}/config.yml | sed -r "s/\s*sonatypeWork\s*:\s*\"?[-0-9a-zA-Z_/\\]
 && chmod 0644 ${CONFIG_HOME}/config.yml
 
 # Create start script
-RUN echo "/usr/bin/java \${JAVA_OPTS} -jar nexus-iq-server-${IQ_SERVER_VERSION}.jar server ${CONFIG_HOME}/config.yml 2> ${LOGS_HOME}/stderr.log" > ${IQ_HOME}/start.sh \
+RUN echo "trap 'kill -TERM \`cut -f1 -d@ $SONATYPE_WORK/lock\`  && sleep 60' SIGTERM" > ${IQ_HOME}/start.sh \
```
mealingr: minor suggestion for consistency/formatting (same for below)
```suggestion
RUN echo "trap 'kill -TERM \`cut -f1 -d@ ${SONATYPE_WORK}/lock\` && sleep 60' SIGTERM" > ${IQ_HOME}/start.sh \
```

mealingr: Rather than a hardcoded 60s, it might be better to wait until the process is stopped with a timeout.
`docker stop ...` will by default only wait for 10s before killing the container
https://docs.docker.com/engine/reference/commandline/stop/
but people can override that with the `-t` option e.g.
`docker stop ... -t 600`
to wait for 10 mins
I believe something like this works although it might be worth extracting the pid to a variable or something
```
RUN echo "trap 'kill -TERM \`cut -f1 -d@ ${SONATYPE_WORK}/lock\`; timeout 600 tail --pid=\`cut -f1 -d@ ${SONATYPE_WORK}/lock\` -f /dev/null' SIGTERM" > ${IQ_HOME}/start.sh \
```
taken from this answer https://stackoverflow.com/a/41613532

The timeout itself could also be a build variable if we wanted instead of another hardcoded 600s (although 10 mins is pretty generous)

The difference between the `sleep 60` and the `timeout 600` above is that the sleep will always happen for that length of time (assuming `docker stop ...` doesn't kill the container), whereas the `timeout 600` will only actually reach 10 mins if the process isn't killed beforehand

srini-hdp: Hi @mealingr 
i've added sleep 60 to add minimum 1 min wait irrespective of the timeout value used with docker -t option or terminateGracePeriod in K8s environment. When higher values are used docker / K8s system would still wait longer
for terminating the pod. This safety net of 1 min, was added to give IQ additional time to close the connections (in case of large env). I've tested with 1.5k apps in DB and scanning against 3 apps when trying to stop IQ. 

We can test the timeout command as well.

anna-d (approved):
+1

Dockerfile
mealingr: Decided to possibly switch to this in a different PR

mealingr (approved):
+1

amiralig:
Merge on behalf of @srini-hdp  as he doesn't seem to have access

PR #72, opened 2021-06-14
Clm 18632 fix iq snapshot

amiralig:
For the full conversation regarding this you can refer to:

https://sonatype.slack.com/archives/CC2BN7KU6/p1622485542189800

We're trying to mount the drive that contains the lock to remove it (sonatype which is mapped on amazon EFS to be accessed by the IQ container) but we're running into a permission issue.

 **This is to clean up some of the discrepancies shown here: https://github.com/sonatype/docker-nexus-iq-server/pull/68. There is no new changes here compared to https://github.com/sonatype/docker-nexus-iq-server/pull/68. this is just to show the net difference between master and this branch**


mealingr (approved):
+1

anna-d (approved):
+1

PR #73, opened 2021-06-16
INT-5317 | Chain RedHat job to Docker Insight job

fblancosona:
Chain the current IQ Docker build to new RedHat publishing job so that it is kicked off automatically

**Links**
JIRA: https://issues.sonatype.org/browse/INT-5317 
Jenkins: https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server-feature/job/INT-5317-chain-redhat-job-insight-job/2/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -178,6 +178,9 @@ node('ubuntu-zion') {
         }
         OsTools.runSafe(this, "git tag -d ${version}")
       }
+
+      //Trigger the Red Hat job and It does not care about this result
+      build(job: '/integrations/Red Hat Certified Docker Image/iq-redhat-certified', wait: false)
     }
 
   } finally {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -180,7 +180,9 @@ node('ubuntu-zion') {
       }
 
       //Trigger the Red Hat job and It does not care about this result
-      build(job: '/integrations/Red Hat Certified Docker Image/iq-redhat-certified', wait: false)
+      stage('Trigger Red Hat Certified Job Build') {
+        build(job: '/integrations/Red Hat Certified Docker Image/iq-redhat-certified', wait: false)
+      }
     }
 
   } finally {


Jenkinsfile
```diff
@@ -178,6 +178,9 @@ node('ubuntu-zion') {
         }
         OsTools.runSafe(this, "git tag -d ${version}")
       }
+
+      //Trigger the Red Hat job and It does not care about this result
+      build(job: '/integrations/Red Hat Certified Docker Image/iq-redhat-certified', wait: false)
```
jflinchbaugh: i think it might be nice to have this wrapped in its own `stage('Run downstream')` or something like that, so it's visible in the UI for build results. 

i think it's fine to have to have this protected by this check for master and push_image, since nothing downstream is prepared to run for anything but master. 👍

jflinchbaugh:
Is there a link to a jenkins build you could add to the description where we can see this having been run?

adrianpowell (approved):

Jenkinsfile
fblancosona: @jflinchbaugh stage added

collinpeters (approved):
+1

collinpeters:
Will give a heads up to the @sonatype/build-release team on this. The tl;dr of it all is that the Redhat bits have been removed from this job and into an Integrations job, and this will trigger that new job from the main Docker job. But it won't wait for the result, should be a fire and forget.

Jenkinsfile
```diff
@@ -178,6 +178,11 @@ node('ubuntu-zion') {
         }
         OsTools.runSafe(this, "git tag -d ${version}")
       }
+
+      //Trigger the Red Hat job and It does not care about this result
+      stage('Trigger Red Hat Certified Job Build') {
+        build(job: '/integrations/Red Hat Certified Docker Image/iq-redhat-certified', wait: false)
```
kellyrob99, line 184: Might also want propagate:false although I'm not certain how that interacts with wait. Pretty sure we don't want this tied to the downstream build result in any case

tneer: Right, if the intent is 'fire and forget' then you don't want the downstream status to propagate back to this job.

https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/#build-build-a-job

PR #75, opened 2021-06-21
BNR-137 Remove obsolete parameter "push_image"

shaikhu:
Jira: https://issues.sonatype.org/browse/BNR-137 

The instructions in [CLM-18731](https://issues.sonatype.org/browse/CLM-18731)  explain, the "push_image" param is used to  enable pushing the image to Docker Hub. 

My assumption is the only time we don't want to push the image to Docker Hub is when the main part of the build passes, but the triggering RH part failed e.g. we wanted to retry publishing to RH, but leaving the Docker Hub image alone.

Now that publishing to RH container registry has been separated, I think it make sense to drop the "push_image" param.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,8 +11,6 @@ properties([
   parameters([
     string(defaultValue: '', description: 'New Nexus IQ Version', name: 'nexus_iq_version'),
     string(defaultValue: '', description: 'New Nexus IQ Version Sha256', name: 'nexus_iq_version_sha'),
-
-    booleanParam(defaultValue: false, description: 'Push Docker Image and Tags', name: 'push_image'),
   ])
 ])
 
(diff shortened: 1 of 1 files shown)


collinpeters (approved):
+1 Had to read the description in BNR-137 to fully understand but this seems reasonable. I was asking myself 'what if we want to build the docker image but not push it to docker hub?' but we don't push it anywhere else so there isn't much point in that

fernau (approved):
LGTM. I like the simplifications. Never used this option, likely no one else did.

shaikhu:
> +1 Had to read the description in BNR-137 to fully understand but this seems reasonable. 

lol sorry about that :sweat_smile: 

> I was asking myself 'what if we want to build the docker image but not push it to docker hub?' but we don't push it anywhere else so there isn't much point in that

I had the same thoughts. Imo when we run this job, it will always be to publish.

PR #77, opened 2021-06-25
add possibility to add certs at runtime to java keystore

kaiehrhardt:


kaiehrhardt:
use the ssl part in the config.yaml

https://help.sonatype.com/iqserver/installing/upgrading-the-iq-server/updating-your-nexus-iq-server-configuration

PR #78, opened 2021-06-26
BNR-141 Compute IQ Version and Sha for docker-nexus-iq-server build

shaikhu:
Jira: https://issues.sonatype.org/browse/BNR-141 
Jenkins Build: https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server/ 

**Description**
Currently the user performing the release has to manually enter the iq version and sha256 (after downloading the iq bundle and using a tool to calculate).

**Current**
image
The user has to manually enter iq version and sha256 (after downloading iq bundle and using an os tool to calculate)

**Proposed Change**
image
User selects the latest iq release from drop down. The version and sha are automatically extracted.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,8 +22,8 @@ node('ubuntu-zion') {
   try {
     stage('Init IQ Version & Sha') {
       nexusIqVersion = getVersionFromBuildName(env.releaseBuild_NAME)
-      gatherBuildArtifacts('insight/insight-brain/release', env.releaseBuild_NUMBER)
-      nexusIqSha = readFile(file: "artifacts/nexus-iq-server-${nexusIqVersion}-bundle.tar.gz.sha256", encoding: 'UTF-8')
+      nexusIqSha = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER,
+        "artifacts/nexus-iq-server-${nexusIqVersion}-bundle.tar.gz.sha256")
     }
     stage('Preparation') {
       deleteDir()
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,15 +7,8 @@
 import com.sonatype.jenkins.pipeline.GitHub
 import com.sonatype.jenkins.pipeline.OsTools
 
-properties([
-  parameters([
-    string(defaultValue: '', description: 'New Nexus IQ Version', name: 'nexus_iq_version'),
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -7,15 +7,8 @@
 import com.sonatype.jenkins.pipeline.GitHub
 import com.sonatype.jenkins.pipeline.OsTools
 
-properties([
```
tneer, line 10: I'm assuming the intent is to replace these with a `run` parameter defined in the UI.

shaikhu: > I'm assuming the intent is to replace these with a `run` parameter defined in the UI.

That's correct. Sorry I forgot to mention that :sweat_smile:.

Jenkinsfile
```diff
@@ -27,6 +20,11 @@ node('ubuntu-zion') {
   GitHub gitHub
 
   try {
+    stage('Init IQ Version & Sha') {
+      nexusIqVersion = getVersionFromBuildName(env.releaseBuild_NAME)
+      nexusIqSha = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER,
```
tneer, line 25: The `readBuildArtifact` method is being added in https://github.com/sonatype/jenkins-shared/pull/140

Jenkinsfile
```diff
@@ -27,6 +20,11 @@ node('ubuntu-zion') {
...
+        "artifacts/nexus-iq-server-${nexusIqVersion}-bundle.tar.gz.sha256")
```
tneer, line 26: Maybe fail the build here if the version and sha aren't available rather than checking for the presence of those variables later.

shaikhu: > Maybe fail the build here if the version and sha aren't available rather than checking for the presence of those variables later.

The `getVersionFromBuildName` library  fails [here](https://github.com/sonatype/jenkins-shared/blob/b4161c5d454d5f2aa2cb129cdb328d06a6f78d84/vars/getVersionFromBuildName.groovy#L24). I've raised a [PR](https://github.com/sonatype/jenkins-shared/pull/140) for reading sha (build artifact). That will also fail [here](https://github.com/sonatype/jenkins-shared/blob/069bf7fd2124b34fd3c4c0c3e38dc23b16ec1abe/vars/readBuildArtifact.groovy#L15). 
Removed checks with 7a7dcc23

tneer (approved):
+1

mealingr (approved):
+1

PR #80, opened 2021-09-21
INT-5698 upgrade and adjust for openjdk-8:1.10 image

jflinchbaugh:
We can't get rid of all the security issues found by artifacthub (with trivy), but upgrading to the base image can chip away at it.

JIRA: https://issues.sonatype.org/browse/INT-5698

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,17 +82,14 @@ RUN cd ${TEMP} \
 && cd ${IQ_HOME} \
 && rm -rf ${TEMP} \
 \
-# Add group and user
+# Add group and user and copy passwd to the nss_wrapper location
 && groupadd -g ${GID} nexus \
 && adduser -u ${UID} -M -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false nexus \
 && cp /etc/passwd /home/jboss/passwd \
+&& chmod 755 /home/jboss \
 \
 # Change owner to nexus user
-&& chown -R nexus:nexus ${IQ_HOME} \
-&& chown -R nexus:nexus ${SONATYPE_WORK} \
-&& chown -R nexus:nexus ${CONFIG_HOME} \
-&& chown -R nexus:nexus ${LOGS_HOME} \
-&& chmod 755 /home/jboss
+&& chown -R nexus:nexus ${IQ_HOME} ${SONATYPE_WORK} ${CONFIG_HOME} ${LOGS_HOME}
 
 # This is where we will store persistent data
 VOLUME ${SONATYPE_WORK}
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,15 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/openjdk-8:1.3-8
+FROM registry.access.redhat.com/ubi8/openjdk-8:1.10-1
 
 # Build parameters
 ARG IQ_SERVER_VERSION=1.124.0-01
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -54,8 +50,9 @@ LABEL name="Nexus IQ Server image" \
 USER root
 
 # For testing
-RUN dnf update --allowerasing -y \
-&& dnf install -y procps
+RUN microdnf update -y \
+&& microdnf install -y procps gzip \
+&& microdnf clean all
```
jflinchbaugh, line 55: switch to microdnf with this new image.

Dockerfile
```diff
@@ -87,13 +84,15 @@ RUN cd ${TEMP} \
 \
 # Add group and user
 && groupadd -g ${GID} nexus \
-&& adduser -u ${UID} -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false nexus \
+&& adduser -u ${UID} -M -d ${IQ_HOME} -c "Nexus IQ user" -g nexus -s /bin/false nexus \
+&& cp /etc/passwd /home/jboss/passwd \
```
jflinchbaugh, line 88: nss_wrapper is configured to find passwd in an alternate location, but adduser added it in the normal place, so copy it. <- this should be a comment in the file i now realize.

Dockerfile
```diff
@@ -85,15 +82,14 @@ RUN cd ${TEMP} \
...
+&& chmod 755 /home/jboss \
 \
 # Change owner to nexus user
-&& chown -R nexus:nexus ${IQ_HOME} \
-&& chown -R nexus:nexus ${SONATYPE_WORK} \
-&& chown -R nexus:nexus ${CONFIG_HOME} \
-&& chown -R nexus:nexus ${LOGS_HOME}
+&& chown -R nexus:nexus ${IQ_HOME} ${SONATYPE_WORK} ${CONFIG_HOME} ${LOGS_HOME}
```
jflinchbaugh, line 92: this could have been done all in one command.

Dockerfile
jflinchbaugh: commented. ✅

jflinchbaugh:
according to trivy, here's the before and after security scans:
before:
```
sonatype/nexus-iq-server:1.124.0 (redhat 8.4)
=============================================
Total: 255 (UNKNOWN: 0, LOW: 103, MEDIUM: 142, HIGH: 6, CRITICAL: 4)
```
after this upgrade:
```
sonatype/nexus-iq-server (redhat 8.4)
=====================================
Total: 160 (UNKNOWN: 0, LOW: 64, MEDIUM: 87, HIGH: 6, CRITICAL: 3)
```

Dockerfile
collinpeters: Can you explain this nss_wrapper a bit more? Googling says this is relating to test functions, but this is our main docker image

jflinchbaugh: i think some older versions of openshift suggest nss_wrapper to allow your app to create and use your own user ids (like our nexus user). UBI8 images are geared toward use on openshift, and that's the image we use for both these images. The latest openjdk-8:1.10 sets it up to this alternate passwd location automatically (see line 295 of https://catalog.redhat.com/software/containers/ubi8/openjdk-8/5dd6a48dbed8bd164a09589a?container-tabs=dockerfile), and chown uses it, but adduser didn't, so I just had to work around it by copying the file we setup. 

Ultimately, we may eventually want to stop using our own UID in kubernetes, but that's a bigger task for another day.

jflinchbaugh: as an alternative, we could strip back the base image to ubi-minimal and install red hat's openjdk like the openjdk-8 image would, just without all the extra stuff: https://github.com/sonatype/docker-nexus-iq-server/pull/82

jflinchbaugh:
preferring #82 instead of this.

PR #85, opened 2021-11-08
CLM-19804 Updated documentation to reflect change to ubi-minimal as base image.

targeter21:


targeter21:
There were several places where whitespace was trimmed. I recommend ignoring whitespace on the diff.

jawadshaikst (approved):
+1

PR #87, opened 2021-11-10
remove trigger of red hat builds

jflinchbaugh:
The plain docker build doesn't need to trigger the broken Red Hat build right now, since the deployment process instead creates a follow-up deployment task in JIRA for the integrations team.

Build: https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server-feature/job/do-not-trigger-red-hat-build/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -175,13 +175,7 @@ node('ubuntu-zion') {
         }
         OsTools.runSafe(this, "git tag -d ${version}")
       }
-
-      //Trigger the Red Hat job and It does not care about this result
-      stage('Trigger Red Hat Certified Job Build') {
-        build(job: '/integrations/Red Hat Certified Docker Image/iq-redhat-certified', wait: false)
-      }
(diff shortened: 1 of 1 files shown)


collinpeters (approved):
+1 - but I thought we had set it up this way on purpose?

fernau (approved):
+1 if this is temporary maybe create a ticket to later enable this automation again.

jflinchbaugh:
> +1 - but I thought we had set it up this way on purpose?

@fblancosona had split it into these 2 builds, but shortly afterward, the red hat apis for triggering the build (used in job 2) changed, so the second job just won't work at this point.

jflinchbaugh:
> +1 if this is temporary maybe create a ticket to later enable this automation again.

added a note to the existing task to restore chaining when the red hat job is fixed: https://issues.sonatype.org/browse/INT-5543

PR #89, opened 2021-12-17
Add git

keeganwitt:
Add Git to make it easy to use native Git instead of JGit (which is a pretty common use case).


maurycupitt:
I would like to see this merged before the next release.

mattjohnson:
The main image now contains git https://github.com/sonatype/docker-nexus-iq-server/blob/main/Dockerfile#L57

PR #93, opened 2022-04-08
INT-6717 alternative slim image

jflinchbaugh:
Additionally produce another docker image that excludes `git` binary and tag it as `<version>-slim` and `latest-slim`, so users can leave out extra tooling if they choose.

JIRA: https://issues.sonatype.org/browse/INT-6717
Build: https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server-feature/job/INT-6717-alternative-slim-image/

Have a look at the build and notice the 2 artifacts, IQ report with 2 occurrences: one for each image, etc. 👍

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,0 +1,116 @@
+# Copyright (c) 2017-present Sonatype, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5-204
+
+# Build parameters
+ARG IQ_SERVER_VERSION=1.135.0-01
+ARG IQ_SERVER_SHA256=5005ba518380ae3dc70f09de5cc43b9e70e7eff979d1fb9ad07ff35985febfdf
+
+
+ARG TEMP="/tmp/work"
+ARG IQ_HOME="/opt/sonatype/nexus-iq-server"
+ARG SONATYPE_WORK="/sonatype-work"
+ARG CONFIG_HOME="/etc/nexus-iq-server"
+ARG LOGS_HOME="/var/log/nexus-iq-server"
+ARG GID=1000
+ARG UID=1000
+ARG TIMEOUT=600
+
+LABEL name="Nexus IQ Server image" \
+  maintainer="Sonatype <support@sonatype.com>" \
+  vendor=Sonatype \
+  version="${IQ_SERVER_VERSION}" \
+  release="1.135.0" \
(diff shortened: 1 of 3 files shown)


Dockerfile.slim
```diff
@@ -0,0 +1,116 @@
...
+  io.k8s.display-name="Nexus IQ Server" \
+  io.openshift.expose-services="8071:8071" \
+  io.openshift.tags="Sonatype,Nexus,IQ Server"
+
+USER root
+
+# For testing
+RUN microdnf update -y \
+&& microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-1.8.0-openjdk-devel \
+&& microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync \
```
jflinchbaugh, line 56: this Dockerfile is a copy of the original Dockerfile, but leaves out `git` here.
Compare to https://github.com/sonatype/docker-nexus-iq-server/blob/INT-6717-alternative-slim-image/Dockerfile#L56

Jenkinsfile
```diff
@@ -97,8 +100,15 @@ node('ubuntu-zion') {
...
       runEvaluation({ stage ->
-        nexusPolicyEvaluation(iqStage: stage, iqApplication: iqApplicationId,
-        iqScanPatterns: [[scanPattern: "container:${imageName}"]], failBuildOnNetworkError: true)}, theStage)
+        nexusPolicyEvaluation(
+          iqStage: stage,
+          iqApplication: iqApplicationId,
+          iqScanPatterns: [
+            [scanPattern: "container:${imageName}"],
+            [scanPattern: "container:${imageName}-slim"],
+          ],
```
eduard-tita, line 109: Question: Have you run this? 

I'm asking because I've never tried to scan two Docker images in the same evaluation. If there are policy violations, most of them will apply to both images. Do we get those violations twice, or are they aggregated? If they are not aggregated, scanning the default image should be enough.

eduard-tita (approved):
+1 with a question/comment

Jenkinsfile
jflinchbaugh: It's aggregated and shows 2 occurrences. Scanning both allows us to see the benefit of the slim image in cases where a default package might be in violation. Have a look at the report linked in the description of the PR. I think it looks pretty intuitive.

nblair: Since it's a separate deliverable, perhaps we should setup a distinct IQ application for the slim image and track it there, rather than combining two into one?

hectorhuol: Also, maybe do we need to scan the RedHat image too?

hectorhuol (approved):
+1 Just adding one comment

Jenkinsfile
jflinchbaugh: We don't build the Red Hat image here at all, only update it. That gets built and scanned in Red Hat's build service. There may be advantage to building it locally and pushing it to their scan service at some point, but I'd reserve another task for that change.

jflinchbaugh: @nblair I thought about making it 2 IDs, but they greatly overlap, so I thought they could share remediations and waivers instead of doubling them. I'm fine either way though. Shall I ask Francisco to create a separate app in IQ , and I'll use 2?

eduard-tita: @nblair If the policy violations are aggregated, I think it's better to keep them under the same IQ app. Otherwise we will get two Jira tickets for each policy violation, but the same change will fix them both. Less overhead for us.

PR #105, opened 2022-10-25
Add Docker stop with timeout command to Readme

hnagrasona:
Customers need to be advised to issue the docker stop with a sufficient timeout, as too often we see avoidable H2 DB corruptions due to abrupt stops of the container.


anna-d (approved):
+1

mattjohnson (approved):

PR #109, opened 2023-01-13
try sign instead of push in the slim image

fvillate:
troubleshooting to figure out if this is the issue related to the docker sign image.


jflinchbaugh:
@fvillate you can probably close this PR.

PR #113, opened 2023-01-13
Upgrade to Java 11

georgenicoll:
Java 1.8 is now very old, upgrade the container images to install Java 11.


vladt:
Some companies have a policy to not use java 11 (some legal issues?).

mattjohnson:
Latest images will begin using OpenJDK 17.

PR #116, opened 2023-01-25
Push_Before_Signing

koraytugay:


Jenkinsfile
```diff
@@ -193,6 +193,12 @@ node('ubuntu-zion-legacy') {
             OsTools.runSafe(this, "docker tag ${slimImageId} ${organization}/${dockerHubRepository}:${version}-slim")
             OsTools.runSafe(this, "docker tag ${slimImageId} ${organization}/${dockerHubRepository}:latest-slim")
 
+            // push the images
+            OsTools.runSafe(this, "docker image push ${organization}/${dockerHubRepository}:${version}")
+            OsTools.runSafe(this, "docker image push ${organization}/${dockerHubRepository}:latest")
+            OsTools.runSafe(this, "docker image push ${organization}/${dockerHubRepository}:${version}-slim")
+            OsTools.runSafe(this, "docker image push ${organization}/${dockerHubRepository}:latest-slim")
+
```
jflinchbaugh: this needs to move down into the withCredentials block below, since DOCKER_CONTENT_TRUST=1 is set above. these will try to sign as you push with that set, so it'll need the PASSPHRASE in the env.

i think the signs will be extra, but that's ok.

alternatively, you could remove DOCKER_CONTENT_TRUST, so it doesn't try to sign with push.

koraytugay: a46ed42cb6f73ad2827d8e818f1af7ca05129bb0

jflinchbaugh (approved):
+1

eduard-tita (approved):
+1

Jenkinsfile
```diff
@@ -196,6 +196,12 @@ node('ubuntu-zion-legacy') {
             // Sign the images
             // Signing images also pushes them
             withCredentials([string(credentialsId: 'sonatype-password', variable: 'DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE')]) {
+              // push the images
+              OsTools.runSafe(this, "docker image push ${organization}/${dockerHubRepository}:${version}")
```
jflinchbaugh, line 200: is it "docker image push" or just "docker push"?

jflinchbaugh: nvm. it is.

PR #120, opened 2023-04-03
POC - Use dockerized build pipelne

shaikhu:
As announced in the Build and Release slack [channel](https://sonatype.slack.com/archives/C01QWPH4HGR/p1680297070189159) the legacy Jenkins agents will be removed soon. The current build doesn't work on the modern Jenkins agent (Ruby is not installed on the modern Jenkins agent).

As mentioned in [BNR-882](https://issues.sonatype.org/browse/BNR-882), this build also has a lot of tech debt. This PR is an initial attempt at modernising this build to use the [dockerizedBuildPipeline](https://docs.sonatype.com/display/CDI/Dockerized+Build+Pipeline).

I've separated the Jenkinsfile into two, one for the main snapshot build, and the second for the release build. While there is some code duplication (the tests), the Jenkinsfiles are easier to read.

Finally, a Jenkinsfile will be needed for each docker file (e.g. slim, redhat etc).

This  POC PR outlines how the build will look using the dockerizedBuildPipeline. Also a place to ask questions or comment if something looks wrong. Also It needs to be fully tested by attempting to publish an image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,6 +51,7 @@ LABEL name="Nexus IQ Server image" \
 USER root
 
 # For testing
+RUN yum install which -y
 RUN microdnf update -y \
 && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-1.8.0-openjdk-devel \
 && microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git \
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ LABEL name="Nexus IQ Server image" \
 USER root
 
 # For testing
-RUN yum install which -y
+RUN microdnf install which -y
 RUN microdnf update -y \
 && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-1.8.0-openjdk-devel \
 && microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git \
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,10 +51,9 @@ LABEL name="Nexus IQ Server image" \
 USER root
 
 # For testing
-RUN microdnf install which -y
 RUN microdnf update -y \
 && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-1.8.0-openjdk-devel \
-&& microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git \
+&& microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git which \
 && microdnf clean all
 
 # Create folders
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,6 +24,7 @@ dockerizedBuildPipeline(
   buildAndTest: {
     // TODO add tests
   },
+  archiveArtifacts: '**/dist/*',
   testResults: ['**/validate-expectations-results.xml'],
   lint: {
     hadolint(['Dockerfile'])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,9 +23,9 @@ dockerizedBuildPipeline(
   },
   buildAndTest: {
     validateExpectations([
-          new Expectation('javaVersion', 'java', '-version', 'openjdk version "1.8.0_342"')
-          /* new Expectation('userGroups', 'id', '', 'uid=1000(nexus) gid=1000(nexus) groups=1000(nexus)'),
-          new Expectation('homeDirectory', 'pwd', '', '/opt/sonatype/nexus-iq-server'),
+          new Expectation('javaVersion', 'java', '-version', 'openjdk version "1.8.0_362"'),
+          new Expectation('userGroups', 'id', '', 'uid=1000(nexus) gid=1000(nexus) groups=1000(nexus)')
+          /*new Expectation('homeDirectory', 'pwd', '', '/opt/sonatype/nexus-iq-server'),
           new Expectation('installDirectory', '/var/log/nexus-iq-server/', '', 'Is a directory'),
           new Expectation('configDirectory', '/etc/nexus-iq-server/', '', 'Is a directory'),
           new Expectation('configFile', 'ls', '/etc/nexus-iq-server', 'config.yml') */
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,12 +23,13 @@ dockerizedBuildPipeline(
   },
   buildAndTest: {
     validateExpectations([
-          new Expectation('javaVersion', 'java', '-version', 'openjdk version "1.8.0_342"'),
-          new Expectation('userGroups', 'id', '', 'uid=1000(nexus) gid=1000(nexus) groups=1000(nexus)'),
+          new Expectation('javaVersion', 'java', '-version', 'openjdk version "1.8.0_342"')
+          /* new Expectation('userGroups', 'id', '', 'uid=1000(nexus) gid=1000(nexus) groups=1000(nexus)'),
           new Expectation('homeDirectory', 'pwd', '', '/opt/sonatype/nexus-iq-server'),
           new Expectation('installDirectory', '/var/log/nexus-iq-server/', '', 'Is a directory'),
           new Expectation('configDirectory', '/etc/nexus-iq-server/', '', 'Is a directory'),
-          new Expectation('configFile', 'ls', '/etc/nexus-iq-server', 'config.yml')])
+          new Expectation('configFile', 'ls', '/etc/nexus-iq-server', 'config.yml') */
+      ])
   },
   testResults: ['**/validate-expectations-results.xml'],
   vulnerabilityScan: {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@Library(['private-pipeline-library', 'jenkins-shared@test-expectations', 'iq-pipeline-library']) _
+@Library(['private-pipeline-library', 'jenkins-shared', 'iq-pipeline-library']) _
 
 import com.sonatype.jenkins.shared.Expectation
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ dockerizedBuildPipeline(
   buildAndTest: {
     validateExpectations([
           new Expectation('javaVersion', 'java', '-version', 'openjdk version "1.8.0_362"'),
-          new Expectation('userGroups', 'id', '', 'uid=1000(nexus) gid=1000(nexus) groups=1000(nexus)')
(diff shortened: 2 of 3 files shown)


Jenkinsfile.release
```diff
@@ -0,0 +1,99 @@
...
+@Library(['private-pipeline-library', 'jenkins-shared']) _
+
+import com.sonatype.jenkins.shared.Expectation
+
+properties([
+  parameters([
+    run(name: 'releaseBuild',
+        filter: 'SUCCESSFUL',
+        projectName: 'insight/insight-brain/release',
+        description: 'The latest release of IQ Server to build the docker image')
```
shaikhu, line 26: This is the release [build](https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/release/) which built and pushed the nexus-iq-server binaries to rso.

Jenkinsfile.release
```diff
@@ -0,0 +1,99 @@
...
+dockerizedBuildPipeline(
+  prepare: {
+    githubStatusUpdate('pending')
+    version = getVersionFromBuildName(env.releaseBuild_NAME)
+    checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
+    updateIQServerVersionAndChecksum(version, checksum)
+    commitAndPushChanges(version)
+  },
+  setVersion: {
+    env['VERSION'] = version.split('-')[0]
```
shaikhu, line 45: For some reason the current build removes the build number from the version string e.g. "1.100.0-01" becomes "1. 100.0". I think we should keep it for consistency (since iq has already been published in docker hub)

Jenkinsfile.release
```diff
@@ -0,0 +1,99 @@
...
+  },
+  buildAndTest: {
+    currentBuild.displayName = "#${currentBuild.id} ${imageName}-${env.VERSION}"
+    // TODO add tests
+  },
+  testResults: ['**/validate-expectations-results.xml'],
+  lint: {
+    hadolint(['Dockerfile'])
```
shaikhu: Any code analysis violations in the Dockerfile will appear as warnings. We can update it to fail the build if needed.

Jenkinsfile.release
```diff
@@ -0,0 +1,99 @@
...
+
+  writeFile(file: "${pwd()}/Dockerfile", text: dockerFile)
+}
+
+void commitAndPushChanges(String version) {
+  runSafely 'git config --global push.default simple'
+  sonatypeZionGitConfig()
+  sshagent(credentials: [sonatypeZionCredentialsId()]) {
+    runSafely 'git add .'
+    runSafely "git diff --exit-code --cached || git commit -m 'Update to version ${version}'"
```
shaikhu, line 101: The `git diff` part ensures if we re-run the build without any changes, the git commit won't be executed (since there's nothing to commit).

Jenkinsfile.release
```diff
@@ -0,0 +1,100 @@
...
+        description: 'The latest release of IQ Server to build the docker image')
+  ])
+])
+
+String imageName = 'sonatype/nexus-iq-server',
+String version = ''
+String checksum = ''
+
+dockerizedBuildPipeline(
+  deployBranch: 'main',
```
shaikhu, line 36: The current default branch is 'master'. We can use this opportunity to rename it to 'main'

Jenkinsfile.release
```diff
@@ -0,0 +1,104 @@
...
+    nexusPolicyEvaluation(
+      iqApplication: 'docker-nexus-iq-server',
+      iqScanPatterns: [[scanPattern: "container:${env.DOCKER_IMAGE_ID}"]],
+      iqStage: 'release')
+  },
+  deploy: {
+    withSonatypeDockerRegistry() {
+      sh """docker tag $DOCKER_IMAGE_ID ${sonatypeDockerRegistryId()}/${imageName}-${env.BUILD_NUMBER}
+            docker push ${sonatypeDockerRegistryId()}/${imageName}-${env.BUILD_NUMBER}
+            docker rmi ${sonatypeDockerRegistryId()}/${imageName}-${env.BUILD_NUMBER}"""
```
shaikhu: This will deploy to our Sonatype Internal Registry. @tneer What needs to be done so we deploy to docker hub instead? Maybe a new shared library e.g. withDockerHub which authenticates against docker hub. The current code is using this Jenkins [credential](https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/credentials/store/folder/domain/_/credential/sonatype-password/).

tneer: @shaikhu I think we need to just copy over and clean up all that stuff from the `pushImages` stage in the original Jenkinsfile.  Most of that is specific to this set of docker images so doesn't really feel like a shared library function to me.

Jenkinsfile
```diff
@@ -13,277 +13,35 @@
...
+
+import com.sonatype.jenkins.shared.Expectation
+
+dockerizedBuildPipeline(
+  lint: {
+    hadolint(['Dockerfile'])
+  },
+  buildAndTest: {
+    validateExpectations([
+          new Expectation('javaVersion', 'java', '-version', 'openjdk version "1.8.0_362"'),
```
tneer: I might suggest a regex for this.  The version may vary slightly over time, which is ok.  For example, 

```groovy
new Expectation('java-version', 'java', '-version', /openjdk version \"11.0.*\" \d{4}-\d{2}-\d{2}/),
```

Jenkinsfile
```diff
@@ -13,277 +13,35 @@
...
+          new Expectation('userGroups', 'id', '', 'uid=1000(nexus) gid=1000(nexus) groups=1000(nexus)'),
```
tneer: I think you want `id nexus` otherwise this is likely to return the info on the jenkins user that is executing the command.

shaikhu: > I think you want `id nexus` otherwise this is likely to return the info on the jenkins user that is executing the command.

Thank you! That fixed it. 143d9e5

Jenkinsfile
shaikhu: > I might suggest a regex for this. The version may vary slightly over time, which is ok. For example,
> 
> ```groovy
> new Expectation('java-version', 'java', '-version', /openjdk version \"11.0.*\" \d{4}-\d{2}-\d{2}/),
> ```

:+1: 2bc190d

Dockerfile
```diff
@@ -51,6 +51,7 @@ LABEL name="Nexus IQ Server image" \
 USER root
 
 # For testing
+RUN microdnf install which -y
```
tneer: Just include this in the list of other packages being installed on line 57.

Jenkinsfile.release
shaikhu: > @shaikhu I think we need to just copy over and clean up all that stuff from the `pushImages` stage in the original Jenkinsfile. Most of that is specific to this set of docker images so doesn't really feel like a shared library function to me.

:+1: a83d1f21

Dockerfile
shaikhu: > Just include this in the list of other packages being installed on line 57.

:+1: a507ec0a

Dockerfile
```diff
@@ -53,7 +53,7 @@ USER root
 # For testing
 RUN microdnf update -y \
 && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-1.8.0-openjdk-devel \
-&& microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git \
+&& microdnf install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git which \
```
eduard-tita, line 56: You probably want to all `which` to the other two Docker files.

Jenkinsfile
```diff
@@ -13,277 +13,34 @@
...
-      return getShortVersion(line.substring(22))
-    }
+@Library(['private-pipeline-library', 'jenkins-shared@test-expectations', 'iq-pipeline-library']) _
+
+import com.sonatype.jenkins.shared.Expectation
+
+dockerizedBuildPipeline(
+  lint: {
+    hadolint(['Dockerfile'])
+  },
```
eduard-tita, line 23: I would not add linting in this PR. It's very likely it fails the build.

Jenkinsfile
```diff
@@ -13,277 +13,34 @@
...
-    stage('Preparation') {
-      deleteDir()
-      OsTools.runSafe(this, "docker system prune -a -f")
-
-      def checkoutDetails = checkout scm
-
-      dockerFileLocations = [
-        "${pwd()}/Dockerfile",
-        "${pwd()}/Dockerfile.slim",
-        "${pwd()}/Dockerfile.rh",
```
eduard-tita, line 53: The original job builds and releases 3 Docker images in the same time (main, slim and RedHat). The new build jobs seem to build and release only the main image (based on `Dockerfile`).

tneer: @eduard-tita That is correct.  We'll need to create a job for each image, which is a bit of a downside, but the upside is that each job will use the standard pipeline which is well supported.

Dockerfile
shaikhu: > You probably want to all `which` to the other two Docker files.

Absolutely :+1: . The complete solution will need `which` installed in each Dockerfile... but this PR is just demonstrating what needs to be done for the main/single docker image :smile:

Jenkinsfile
shaikhu: > I would not add linting in this PR. It's very likely it fails the build.

Yes, correct again. For example see this[ build](https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server-feature/job/poc_dockerized_build_pipeline/37/). Even though the lint stage failed, the remaining stages ran to completion, although the final build result is marked as failed. According to the dockerizedBuildPipeline [docs](https://docs.sonatype.com/pages/viewpage.action?spaceKey=CDI&title=Dockerized+Build+Pipeline#DockerizedBuildPipeline-lint) this stage is optional and can be left out.

Jenkinsfile.release
eduard-tita: This is used as a Docker tag. Now this images has tags like `1.158.0` in DockerHub. The slim image tags look like `1.158.0-slim`. I guess we want to keep this format going forward.

shaikhu: > This is used as a Docker tag. Now this images has tags like `1.158.0` in DockerHub. The slim image tags look like `1.158.0-slim`. I guess we want to keep this format going forward.

Sorry I realised my above [comment](https://github.com/sonatype/docker-nexus-iq-server/pull/120#discussion_r1155802392) was incorrect. I've updated it. Yes you're correct the docker tag will use the same format it has been using e.g. 1.100.0.

shaikhu:
I think this PR has gotten enough views/comments from the right people, so going to close it. I think the @sonatype/build-release team will reach out to individual teams the closer we get to the deadline for getting rid of the legacy jenkins agent. Hopefully this PR gives an outline what to do for using the dockerizedBuildPipeline (which will work on the modern jenkins agent).

PR #123, opened 2023-05-22
CLM-22746: Refactor Jenkins Pipeline

agustin-mrtz:


Jenkinsfile
```diff
@@ -14,276 +14,50 @@
...
-  def content = readFile 'Dockerfile'
-  for (line in content.split('\n')) {
-    if (line.startsWith('ARG IQ_SERVER_VERSION=')) {
-      return getShortVersion(line.substring(22))
-    }
+import com.sonatype.jenkins.shared.Expectation
+
+dockerizedBuildPipeline(
+  prepare: {
+    githubStatusUpdate('pending')
```
shaikhu, line 22: The Jenkinsfile refers to the main-snapshot build right? IMO it might be overboard updating the github status for that build.

Jenkinsfile
```diff
@@ -14,276 +14,50 @@
...
+  },
+  lint: {
+    hadolint(['./Dockerfile'])
+  },
+  setVersion: {
+    env['VERSION'] = "${env.BRANCH_NAME}-${env.BUILD_NUMBER}"
+  },
```
shaikhu: The version environment variable doesn't seem to be used in this file. I think you can drop setting it.

Jenkinsfile
```diff
@@ -14,276 +14,50 @@
...
+      iqApplication: 'docker-nexus-iq-server',
+      iqScanPatterns: [[scanPattern: "container:${env.DOCKER_IMAGE_ID}"]],
+      iqStage: 'develop')
+  },
+  onSuccess: {
+    githubStatusUpdate('success')
+  },
+  onFailure: {
+    githubStatusUpdate('failure')
   }
```
shaikhu: For the main snapshot build, rather then updating the github status , I think it would be better to send a notification (for both build success/failure) to the iq-builds slack channel.  This can be done with the code 
```
buildNotifications(currentBuild, env)
```
See this example slack [message](https://sonatype.slack.com/archives/CCA312MCZ/p1684824996982179) for the insight-brain build.

Jenkinsfile.release
```diff
@@ -0,0 +1,175 @@
...
+  ])
+])
+
+String imageName = 'sonatype/nexus-iq-server',
+String dockerHubRepository = 'nexus-iq-server',
+String version = ''
+String checksum = ''
+
+dockerizedBuildPipeline(
+  deployBranch: 'main',
```
shaikhu, line 35: Just a reminder once this PR is merged, you (or somebody from the B&R team can help) need to rename the default branch name from `master` to `main` in Github branch settings.

Jenkinsfile.slim
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@Library(['private-pipeline-library', 'jenkins-shared']) _
+import com.sonatype.jenkins.shared.Expectation
+
+dockerizedBuildPipeline(
+  prepare: {
+    githubStatusUpdate('pending')
```
shaikhu, line 21: Can probably drop this for a snapshot build

Jenkinsfile.slim
```diff
@@ -0,0 +1,64 @@
...
+  },
+  setVersion: {
+    env['VERSION'] = "${env.BRANCH_NAME}-${env.BUILD_NUMBER}"
+  },
```
shaikhu: version environment doesn't seem to be used elsewhere. Can probably drop this block.

Jenkinsfile.slim
```diff
@@ -0,0 +1,64 @@
...
+      iqApplication: 'docker-nexus-iq-server',
+      iqScanPatterns: [[scanPattern: "container:${env.DOCKER_IMAGE_ID}"]],
+      iqStage: 'develop')
+  },
+  onSuccess: {
+    githubStatusUpdate('success')
+  },
+  onFailure: {
+    githubStatusUpdate('failure')
+  }
```
shaikhu: If we want to keep this snapshot build, probably better to send a notification to the iq-builds slack channel instead. See [comment](https://github.com/sonatype/docker-nexus-iq-server/pull/123/files#r1201937484)

Jenkinsfile.rh
```diff
@@ -14,46 +14,51 @@
...
-    sh 'docker logout'
-    sh 'docker system prune -a -f'
-    sh 'git clean -f && git reset --hard origin/master'
+dockerizedBuildPipeline(
+  prepare: {
+    githubStatusUpdate('pending')
+  },
+  setVersion: {
+    env['VERSION'] = "${env.BRANCH_NAME}-${env.BUILD_NUMBER}"
+  },
```
shaikhu: Can drop for the snapshot build

Jenkinsfile.rh
```diff
@@ -14,46 +14,51 @@
...
+      iqApplication: 'docker-nexus-iq-server',
+      iqScanPatterns: [[scanPattern: "container:${env.DOCKER_IMAGE_ID}"]],
+      iqStage: 'develop')
+  },
+  onSuccess: {
+    githubStatusUpdate('success')
+  },
+  onFailure: {
+    githubStatusUpdate('failure')
   }
```
shaikhu, line 63: same https://github.com/sonatype/docker-nexus-iq-server/pull/123/files#r1201937484

Jenkinsfile
tneer: I think it can be any branch.  I would probably leave this here unless the GitHub checks are sent by Jenkins as part of the job definition (typically only GitHub branch source jobs).  These status checks can be used for branch protection rules.

Jenkinsfile
```diff
@@ -14,276 +14,50 @@
...
+    nexusPolicyEvaluation(
+      iqApplication: 'docker-nexus-iq-server',
+      iqScanPatterns: [[scanPattern: "container:${env.DOCKER_IMAGE_ID}"]],
+      iqStage: 'develop')
+  },
+  onSuccess: {
+    githubStatusUpdate('success')
+  },
+  onFailure: {
+    githubStatusUpdate('failure')
```
tneer, line 45: You might want to also notify failures to a Slack channel.

Like https://github.com/sonatype/jenkins-shared-library-support/blob/be88e227f08a131ac9c23b116a99bd266319b757/Jenkinsfile.docker#L54

You can wrap that with something like `if (env.BRANCH_NAME == deployBranch) {` if you only want to notify of failures on the main branc.

Jenkinsfile
tneer: I would agree that adding a slack message is a good idea, `buildNotifications` is in the `iq-pipeline-library` and sends email and slack messages.  See my other note on just sending a slack message.  I do think you should keep the GitHub status updates unless you are using the GitHub branch source in your job config which also sends them.

Jenkinsfile
```diff
@@ -14,276 +14,50 @@
...
+  setVersion: {
+    env['VERSION'] = "${env.BRANCH_NAME}-${env.BUILD_NUMBER}"
+  },
+  buildAndTest: {
+    validateExpectations([
+        new Expectation('nexus-group', 'grep', '^nexus: /etc/group', 'nexus:x:1000:'),
+        new Expectation('nexus-user', 'grep', '^nexus: /etc/passwd', 'nexus:x:1000:1000:Nexus IQ user:/opt/sonatype/nexus-iq-server:/bin/false'),
+        new Expectation('iq-process', 'ps', '-e -o command,user | grep -q ^/usr/bin/java.*nexus$ | echo $?', '0'),
+        new Expectation('application-port', 'curl', '-s --fail --connect-timeout 120 http://localhost:8070/ | echo $?', '0'),
+        new Expectation('admin-port', 'curl', '-s --fail --connect-timeout 120 http://localhost:8070/ | echo $?', '0'),
```
mealingr: (same for the other files)
```suggestion
        new Expectation('admin-port', 'curl', '-s --fail --connect-timeout 120 http://localhost:8071/ | echo $?', '0'),
```

Jenkinsfile
agustin-mrtz: This file and all other Jenkinsfile that don't end in release are for feature branches

Jenkinsfile.release
```diff
@@ -0,0 +1,175 @@
...
+      [ $class: 'UsernamePasswordMultiBinding', credentialsId: 'docker-hub-credentials',
+        usernameVariable: 'DOCKERHUB_API_USERNAME', passwordVariable: 'DOCKERHUB_API_PASSWORD']]) {
+
+      runSafely """docker login --username ${env.DOCKERHUB_API_USERNAME} --password ${env.DOCKERHUB_API_PASSWORD}"
+                   docker trust key load $NEXUS_IQ_SERVER_REPOSITORY_KEY
+                   docker trust key load $SONATYPE_KEY"""
+
+      // add signer - for this you need signers public key and repository keys password
+      withCredentials([string(credentialsId: 'nexus-iq-server_dct_reg_pw', variable: 'DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE')]) {
+        runSafely "docker trust signer add sonatype ${organization}/${dockerHubRepository} --key $SONATYPE_PUB"
```
kellyrob99: note that your $org/$repo here is already declared as the `imageName` variable above, can probably clean that up a bit

Jenkinsfile.release
agustin-mrtz: Yep, we want to keep both until we are satisfied with testing of this change

Jenkinsfile
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/pull/123/commits/78a503dbcd229917ff3cb39b0464d42f9a68d9e2

Jenkinsfile.slim
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/pull/123/commits/78a503dbcd229917ff3cb39b0464d42f9a68d9e2

Jenkinsfile.slim
agustin-mrtz: I don't think we want to keep the feature branch build. I think it will be used more as a verification step before merging to main

Jenkinsfile.rh
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/pull/123/commits/78a503dbcd229917ff3cb39b0464d42f9a68d9e2

Jenkinsfile
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/pull/123/commits/78a503dbcd229917ff3cb39b0464d42f9a68d9e2

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+            git tag ${env.VERSION}
+            git push origin ${env.VERSION}'''
+    }
+  },
+  onSuccess: {
+    githubStatusUpdate('success')
+  },
+  onFailure: {
+    githubStatusUpdate('failure')
+    notifyChat(currentBuild: currentBuild, env: env, room: 'iq-builds')
```
shaikhu: Sorry for any confusion, when I suggested adding a notification, it was for the main snapshot build, not the release build. So this would need to be in the Jenkinsfile, not the Jenkinsfile.release file. 

By sending the notification for main snapshot build (which will scheduled on a timer), developers can know the overall health of the main branch). See Troy's [comment](https://github.com/sonatype/docker-nexus-iq-server/pull/123#discussion_r1202513207) on how to send notifications just for `main` and not all feature branches e.g.

```
if (env.BRANCH_NAME == deployBranch)) {
    notifyChat(currentBuild: currentBuild, env: env, room: 'iq-builds')
}
```

The release build is very rarely ran by it's own, and normally part of the IQ Release pipeline (which will send notifications if that fails). I don't see much benefit in sending a notification here.

shaikhu (approved):
+1 with one [comment](https://github.com/sonatype/docker-nexus-iq-server/pull/123/files#r1203786922)

Jenkinsfile
```diff
@@ -14,276 +14,47 @@
...
+        new Expectation('request-log', 'test', '-f /var/log/nexus-iq-server/request.log | echo $?', '0'),
+        new Expectation('stderr-log', 'test', '-f /var/log/nexus-iq-server/stderr.log | echo $?', '0'),
+        new Expectation('home-directory', 'ls', '-la /opt/sonatype | grep nexus-iq-server | awk \'\$9 !~ /^\\.*$/{print \$1,\$3,\$4,\$9}\'', 'drwxr-xr-x nexus nexus nexus-iq-server'),
+        new Expectation('start-script', 'test', '-f /opt/sonatype/nexus-iq-server/start.sh | echo $?', '0'),
+        new Expectation('start-script-has-java-opts', 'grep', '\'JAVA_OPTS\' /opt/sonatype/nexus-iq-server/start.sh | echo $?', '0'), 
+        new Expectation('work-directory', 'ls', '-la / | grep sonatype-work | awk \'\$9 !~ /^\\.*$/{print \$1,\$3,\$4,\$9}\'', 'drwxr-xr-x nexus nexus sonatype-work'), 
+        new Expectation('data-directory', 'test', '-d /sonatype-work/data | echo $?', '0'), 
+        new Expectation('config-directory', 'ls', '-la /etc | grep nexus-iq-server | awk \'\$9 !~ /^\\.*$/{print \$1,\$3,\$4,\$9}\'', 'drwxr-xr-x nexus nexus nexus-iq-server'), 
+        new Expectation('config-file', 'test', '-f /etc/nexus-iq-server | echo $?', '0')
```
mealingr: Wondering if we could extract these checks into a separate file and import them since they're same across all the images

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+  parameters([
+    run(name: 'releaseBuild',
+        filter: 'SUCCESSFUL',
+        projectName: 'insight/insight-brain/release',
+        description: 'The latest release of IQ Server to build the docker image')
+  ])
+])
+
+String imageName = 'sonatype/nexus-iq-server',
+String dockerHubRepository = 'nexus-iq-server',
```
mealingr: Can we remove these commas?

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+  }
+)
+
+void updateIQServerVersionAndChecksum(String version, String checksum) {
+  def dockerFile = readFile(file: "${pwd()}/Dockerfile")
+  def versionRegex = /(ARG IQ_SERVER_VERSION=)(\d\.\d{1,3}\.\d\-\d{2})/
+  def shaRegex = /(ARG IQ_SERVER_SHA256=)([A-Fa-f0-9]{64})/
+
+  dockerFile = dockerFile.replaceAll(versionRegex, "\$1${version}")
+  dockerFile = dockerFile.replaceAll(shaRegex, "\$1${checksum}")
```
mealingr, line 93: In the original Jenkinsfile we were also replacing the `metaShortVersionRegex`, should we also do that here?

https://github.com/sonatype/docker-nexus-iq-server/blob/master/Jenkinsfile#L283-L284

https://github.com/sonatype/docker-nexus-iq-server/blob/master/Dockerfile#L35

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+
+  writeFile(file: "${pwd()}/Dockerfile", text: dockerFile)
+}
+
+void commitAndPushChanges(String version) {
+  runSafely 'git config --global push.default simple'
+  sonatypeZionGitConfig()
+  sshagent(credentials: [sonatypeZionCredentialsId()]) {
+    runSafely 'git add .'
+    runSafely "git diff --exit-code --cached || git commit -m 'Update to version ${version}'"
```
mealingr: (to match https://github.com/sonatype/docker-nexus-iq-server/blob/master/Jenkinsfile#L139-L140)
```suggestion
    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ Server to ${version}.'"
```

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+      iqStage: 'release')
+  },
+  deploy: {
+    pushImage()
+  },
+  postDeploy: {
+    sshagent(credentials: [sonatypeZionCredentialsId()]) {
+      sh '''git config user.email "sonatype-zion@sonatype.com"
+            git tag ${env.VERSION}
+            git push origin ${env.VERSION}'''
```
mealingr, line 74: It looks like we were also deleting the tag locally after pushing it
https://github.com/sonatype/docker-nexus-iq-server/blob/master/Jenkinsfile#L236

but I'm not sure if that's needed

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+
+dockerizedBuildPipeline(
+  deployBranch: 'main',
+  prepare: {
+    githubStatusUpdate('pending')
+    version = getVersionFromBuildName(env.releaseBuild_NAME)
+    checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
+    updateIQServerVersionAndChecksum(version, checksum)
+    commitAndPushChanges(version)
+  },
```
mealingr, line 42: Is this order different to previously?

IIUC previously we were committing and pushing the version updates after building/testing/evaluating but here it looks like we're committing and pushing the version updates before building/testing/evaluating?

One reason this might matter is in the past we've had to re-run the release because of evaluation failures
https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server/
so if we commit/push before we might end up with some unnecessary commits

On the other hand this would help avoid the issue of commits to main whilst a release is happening.

Not sure if we want to change the process as part of this though.

Jenkinsfile.release
```diff
@@ -0,0 +1,176 @@
...
+])
+
+String imageName = 'sonatype/nexus-iq-server',
+String dockerHubRepository = 'nexus-iq-server',
+String version = ''
+String checksum = ''
+
+dockerizedBuildPipeline(
+  deployBranch: 'main',
+  prepare: {
```
mealingr, line 36: Maybe more of a question for the B&R team

Previously we were running `docker system prune -a -f` to cleanup unused docker resources

Do we still need to do that or is this taken care of with the new pipeline?

Jenkinsfile.slim.release
```diff
@@ -0,0 +1,178 @@
...
+  parameters([
+    run(name: 'releaseBuild',
+        filter: 'SUCCESSFUL',
+        projectName: 'insight/insight-brain/release',
+        description: 'The latest release of IQ Server to build the docker image')
+  ])
+])
+
+String imageName = 'sonatype/nexus-iq-server-slim',
+String slimImageId = 'sonatype/nexus-iq-server-slim',
```
mealingr: Can get rid of one of these I think (both the same)

Jenkinsfile.release
shaikhu: > Maybe more of a question for the B&R team
> 
> Previously we were running `docker system prune -a -f` to cleanup unused docker resources
> 
> Do we still need to do that or is this taken care of with the new pipeline?

@mealingr Yes, the `dockerizedBuildPipeline` will call `dockerRemoveImages` [here](https://github.com/sonatype/jenkins-shared/blob/1ba07114935125cb2a14f3bebf08e40c97864bc7/vars/dockerizedBuildPipeline.groovy#L133). From the [docs](https://docs.sonatype.com/display/CDI/dockerRemoveImages) _"Remove stale docker images using docker image prune, but do not fail the build if there are errors."_

mealingr: Thanks @shaikhu 👍

Jenkinsfile.release
agustin-mrtz: You're right, this version doesn't follow the exact order of steps. I can do this step in the `postDeploy` stage if appropriate.

Dockerfile
```diff
@@ -74,9 +81,9 @@ RUN echo "trap 'kill -TERM \`cut -f1 -d@ ${SONATYPE_WORK}/lock\`; timeout ${TIME
 && echo "wait" >> ${IQ_HOME}/start.sh \
 && chmod 0755 ${IQ_HOME}/start.sh
 
+WORKDIR ${TEMP}
 # Download the server bundle, verify its checksum, and extract the server jar to the install directory
-RUN cd ${TEMP} \
-&& curl -L https://download.sonatype.com/clm/server/nexus-iq-server-${IQ_SERVER_VERSION}-bundle.tar.gz --output nexus-iq-server-${IQ_SERVER_VERSION}-bundle.tar.gz \
+RUN curl -L https://download.sonatype.com/clm/server/nexus-iq-server-${IQ_SERVER_VERSION}-bundle.tar.gz --output nexus-iq-server-${IQ_SERVER_VERSION}-bundle.tar.gz \
```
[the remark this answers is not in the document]
agustin-mrtz: Fixed

Dockerfile
```diff
@@ -14,7 +14,8 @@
 # limitations under the License.
 #
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+# hadolint ignore=DL3026
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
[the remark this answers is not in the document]
agustin-mrtz: Fixed

Jenkinsfile.slim.release
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.release#L30

Jenkinsfile.release
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.release#L105

Jenkinsfile.release
agustin-mrtz: Added: https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.release#L89

Jenkinsfile.release
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.release#L30

Jenkinsfile
agustin-mrtz: Addressed: 
https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/expectations.groovy 

https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile#L27

Jenkinsfile.release
agustin-mrtz: Addressed: https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.release#L138

Jenkinsfile.slim.release
```diff
@@ -0,0 +1,162 @@
...
+
+  writeFile(file: "${pwd()}/Dockerfile.slim", text: dockerFile)
+}
+
+void commitAndPushChanges(String version) {
+  runSafely 'git config --global push.default simple'
+  sonatypeZionGitConfig()
+  sshagent(credentials: [sonatypeZionCredentialsId()]) {
+    runSafely 'git add .'
+    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ version to ${version}.'"
```
mealingr:
```suggestion
    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ Server to ${version}.'"
```

Jenkinsfile.release
mealingr: > You're right, this version doesn't follow the exact order of steps. I can do this step in the `postDeploy` stage if appropriate.

I would do this if possible.

i.e. I would avoid changing the order as part of this PR and try to default to how it was working previously. Changing the order may be something we have to discuss in a larger context

shaikhu: > > You're right, this version doesn't follow the exact order of steps. I can do this step in the `postDeploy` stage if appropriate.
> 
> I would do this if possible.
> 
> i.e. I would avoid changing the order as part of this PR and try to default to how it was working previously. Changing the order may be something we have to discuss in a larger context

@mealingr @agustin-mrtz I don't think that moving this to the `postDeploy` stage will work, as the version and checksum in the dockerfile needs to be updated **before** the image is built and pushed to docker hub e.g. before the deploy. 

You're correct that the commit now happens before the building/test/evaluation but I think (this really needs to be tested) that we should be ok, even if the an evaluation fails and we need to rerun the build (with the caveat the same release number is used). 

The following [code](https://github.com/sonatype/docker-nexus-iq-server/blob/b1722861de1577da808fe846dcdbd6dc8dfb8cdd/Jenkinsfile.release#L105) ` runSafely "git diff --exit-code --cached || git commit -m 'Update IQ version to ${version}.'"` will only commit if an actual code change is detected. So if the same release number is used, no change would be detected and a commit shouldn't happen.

mealingr: > The following [code](https://github.com/sonatype/docker-nexus-iq-server/blob/b1722861de1577da808fe846dcdbd6dc8dfb8cdd/Jenkinsfile.release#L105)  runSafely "git diff --exit-code --cached || git commit -m 'Update IQ version to ${version}.'" will only commit if an actual code change is detected. So if the same release number is used, no change would be detected and a commit shouldn't happen.

Nice spot, so if we specify the exact same IQ Server version, it won't make/push a commit.

Also if the build/test/evaluation fails it won't have pushed the image so we shouldn't have a problem there either.

The tagging should be good too, it may not always point to the "Update IQ Server to X." commit (if the release is reran but someone committed to main before the rerun) but I think that's fine.

I agree that we should be ok (no changes needed @agustin-mrtz ) and can give this a try.

Thanks @shaikhu 👍

mealingr (approved):
+1

Jenkinsfile.release
agustin-mrtz: I don't think we need to delete the tags.

Jenkinsfile.release
agustin-mrtz: Agreed, I corrected the files.

Notification sent if branch is deploy branch: 
https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile 
https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.rh 
https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.slim

Only status updates made for release files: 
https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.release 
https://github.com/sonatype/docker-nexus-iq-server/blob/CLM-22746_update_builder_node/Jenkinsfile.slim.release

PR #137, opened 2023-07-31
Remove obsolete/unused Jenkinsfile

shaikhu:
Updating the [readme](https://github.com/sonatype/docker-nexus-iq-server/blob/main/README.md) was moved into the main docker release [jenkinsfile](https://github.com/sonatype/docker-nexus-iq-server/blob/8374b44d5f336a53836259cc6e5f5b61718f0d23/Jenkinsfile.release#L153). It seems the separate Jenkinsfile.updateReadme was not removed.

This PR removes the obsolete Jenkinsfile.updateReadme file as the [build](https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/docker-nexus-iq-server-update-readme/) hasn't been used in 3 years.


tneer (approved):

kellyrob99:
@shaikhu looks like the code to perform this is still being done 3X since it got pulled into each of the 3 image releases. In theory it's all the same content posted in triplicate, but that's probably not intentional.
@agustin-mrtz @koraytugay what do you guys think?

koraytugay:
Yes, @kellyrob99 but only 2 I think not 3.. redhat image does not have the readme. Pushing readme can be removed from docker hub slim release file without IMO.

shaikhu:
> @shaikhu looks like the code to perform this is still being done 3X since it got pulled into each of the 3 image releases. In theory it's all the same content posted in triplicate, but that's probably not intentional. @agustin-mrtz @koraytugay what do you guys think?

Agreed. Removed from the slim-image release with 33fcb98

shaikhu:
> Yes, @kellyrob99 but only 2 I think not 3.. redhat image does not have the readme. Pushing readme can be removed from docker hub slim release file without IMO.

Yes only 2, the main and slim image

PR #140, opened 2023-08-04
CLM-26854: Update Dockerfile.rh with IQ Version and checksum

johnatan-velez:
Jira: https://sonatype.atlassian.net/browse/CLM-26854

Include instructions in the Jenkins.rh.release file to update the Docker.rh file with the IQ version and checksum during the release process.

### DIFF
--- a/Jenkinsfile.rh.release
+++ b/Jenkinsfile.rh.release
@@ -56,7 +56,7 @@ void commitAndPushChanges(String version) {
   sonatypeZionGitConfig()
   sshagent(credentials: [sonatypeZionCredentialsId()]) {
     runSafely 'git add .'
-    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ Server to ${version}.'"
+    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ Server to ${version} for the Red Hat image.'"
 
     // pull and merge any new commits on main so that the push doesn't fail
     runSafely 'git pull --no-rebase --no-edit origin main'
--- a/Jenkinsfile.rh.release
+++ b/Jenkinsfile.rh.release
@@ -24,7 +24,6 @@ properties([
   ]),
 ])
 
-String imageName = 'sonatype/nexus-iq-server'
 String version = ''
 String checksum = ''
 
--- a/Jenkinsfile.rh.release
+++ b/Jenkinsfile.rh.release
@@ -27,17 +27,6 @@ properties([
 String version = ''
 String checksum = ''
 
-dockerizedBuildPipeline(
-  deployBranch: 'main',
-  prepare: {
-    githubStatusUpdate('pending')
-    version = getVersionFromBuildName(env.releaseBuild_NAME)
-    checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
-    updateIQServerVersionAndChecksum(version, checksum)
-    commitAndPushChanges(version)
-  }
-)
-
 void updateIQServerVersionAndChecksum(String version, String checksum) {
   def dockerFile = readFile(file: 'Dockerfile.rh')
   def metaShortVersionRegex = /(release=")(\d\.\d{1,3}\.\d)(" \\)/
@@ -67,6 +56,19 @@ void commitAndPushChanges(String version) {
 node('ubuntu-zion') {
   try {
     stage('Preparation') {
+      try {
+        githubStatusUpdate('pending')
+        version = getVersionFromBuildName(env.releaseBuild_NAME)
+        checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
+        updateIQServerVersionAndChecksum(version, checksum)
+        commitAndPushChanges(version)
+        githubStatusUpdate('success')
+      }
+      catch (Exception e) {
+        githubStatusUpdate('failure')
+        throw e
+      }
+
       deleteDir()
 
       checkout scm
--- a/Jenkinsfile.rh.release
+++ b/Jenkinsfile.rh.release
@@ -48,7 +48,7 @@ void updateIQServerVersionAndChecksum(String version, String checksum) {
   dockerFile = dockerFile.replaceAll(versionRegex, "\$1${version}")
   dockerFile = dockerFile.replaceAll(shaRegex, "\$1${checksum}")
 
-  writeFile(file: 'Dockerfile.rh', text: dockerFile.rh)
+  writeFile(file: 'Dockerfile.rh', text: dockerFile)
 }
 
 void commitAndPushChanges(String version) {
--- a/Jenkinsfile.rh.release
+++ b/Jenkinsfile.rh.release
@@ -27,47 +27,11 @@ properties([
 String version = ''
 String checksum = ''
 
-void updateIQServerVersionAndChecksum(String version, String checksum) {
-  def dockerFile = readFile(file: 'Dockerfile.rh')
-  def metaShortVersionRegex = /(release=")(\d\.\d{1,3}\.\d)(" \\)/
-  def versionRegex = /(ARG IQ_SERVER_VERSION=)(\d\.\d{1,3}\.\d\-\d{2})/
-  def shaRegex = /(ARG IQ_SERVER_SHA256=)([A-Fa-f0-9]{64})/
-  dockerFile = dockerFile.replaceAll(metaShortVersionRegex,
-      "\$1${version.substring(0, version.indexOf('-'))}\$3")
-  dockerFile = dockerFile.replaceAll(versionRegex, "\$1${version}")
-  dockerFile = dockerFile.replaceAll(shaRegex, "\$1${checksum}")
-
-  writeFile(file: 'Dockerfile.rh', text: dockerFile)
-}
-
-void commitAndPushChanges(String version) {
-  runSafely 'git config --global push.default simple'
-  sonatypeZionGitConfig()
-  sshagent(credentials: [sonatypeZionCredentialsId()]) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile.rh.release
```diff
@@ -24,6 +24,46 @@ properties([
   ]),
 ])
 
+String version = ''
+String checksum = ''
+
+dockerizedBuildPipeline(
+  deployBranch: 'main',
+  prepare: {
```
agustin-mrtz: There is a prepare stage already I think you can put this stuff there instead of declaring another pipeline

Jenkinsfile.rh.release
```diff
@@ -24,6 +24,46 @@ properties([
...
+    githubStatusUpdate('pending')
+    version = getVersionFromBuildName(env.releaseBuild_NAME)
+    checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
+    updateIQServerVersionAndChecksum(version, checksum)
+    commitAndPushChanges(version)
+  }
+)
+
+void updateIQServerVersionAndChecksum(String version, String checksum) {
```
agustin-mrtz: Looks correct - it will be the same as https://github.com/sonatype/docker-nexus-iq-server/blob/master/Jenkinsfile#L275

agustin-mrtz:
I understand why this was needed now and how it was missed during the migration. Previously - the Dockerfiles were all updated at once since they were built in the same pipeline. This looks like the correct approach/fix.

Jenkinsfile.rh.release
```diff
@@ -24,6 +24,46 @@ properties([
...
+
+  writeFile(file: 'Dockerfile.rh', text: dockerFile.rh)
+}
+
+void commitAndPushChanges(String version) {
+  runSafely 'git config --global push.default simple'
+  sonatypeZionGitConfig()
+  sshagent(credentials: [sonatypeZionCredentialsId()]) {
+    runSafely 'git add .'
+    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ Server to ${version}.'"
```
kakumara: With this we are committing the Dockerfile.rh file update, right? But the message is the same as https://github.com/sonatype/docker-nexus-iq-server/blob/main/Jenkinsfile.release#L103 . Shouldn't we differentiate a bit to see this clearly?

kakumara (approved):
+1 (with a comment)

Jenkinsfile.rh.release
johnatan-velez: Commit: 6c5f626c8bc085c2cc95e48fe6cbbb4e6b0430a7

Jenkinsfile.rh.release
```diff
@@ -24,6 +24,46 @@ properties([
...
+
+  writeFile(file: 'Dockerfile.rh', text: dockerFile.rh)
+}
+
+void commitAndPushChanges(String version) {
+  runSafely 'git config --global push.default simple'
+  sonatypeZionGitConfig()
+  sshagent(credentials: [sonatypeZionCredentialsId()]) {
+    runSafely 'git add .'
+    runSafely "git diff --exit-code --cached || git commit -m 'Update IQ Server to ${version} for the Red Hat image.'"
```
mr-zepol: Since we have a new branding, should we say  `Sonatype IQ Server` instead?

agustin-mrtz: I think that should be a separate story. There are two other images we release. If we update branding for one image we should do it for the rest as well.

johnatan-velez: Anyways, since this is just a commit message, and internal, I wouldn't worry about that.

Jenkinsfile.rh.release
johnatan-velez: Commit: 8912bac3a158efe5943e9c752169ca611c7d4cc5

Jenkinsfile.rh.release
```diff
@@ -24,9 +24,44 @@ properties([
...
+  def dockerFile = readFile(file: 'Dockerfile.rh')
+  def metaShortVersionRegex = /(release=")(\d\.\d{1,3}\.\d)(" \\)/
+  def versionRegex = /(ARG IQ_SERVER_VERSION=)(\d\.\d{1,3}\.\d\-\d{2})/
+  def shaRegex = /(ARG IQ_SERVER_SHA256=)([A-Fa-f0-9]{64})/
+  dockerFile = dockerFile.replaceAll(metaShortVersionRegex,
+      "\$1${version.substring(0, version.indexOf('-'))}\$3")
+  dockerFile = dockerFile.replaceAll(versionRegex, "\$1${version}")
+  dockerFile = dockerFile.replaceAll(shaRegex, "\$1${checksum}")
+
+  writeFile(file: 'Dockerfile.rh', text: dockerFile.rh)
```
agustin-mrtz: `writeFile(file: 'Dockerfile.rh', text: dockerFile)`

Jenkinsfile.rh.release
```diff
@@ -24,9 +24,44 @@ properties([
...
+    // pull and merge any new commits on main so that the push doesn't fail
+    runSafely 'git pull --no-rebase --no-edit origin main'
+    runSafely 'git push origin HEAD:main'
+  }
+}
+
 node('ubuntu-zion') {
   try {
     stage('Preparation') {
+      githubStatusUpdate('pending')
```
agustin-mrtz: If we are going to add this status we should add more statuses to indicate completion or else it'll stay pending forever

agustin-mrtz: the dockerized pipeline does this: 
```
  onSuccess: {
    githubStatusUpdate('success')
  },
  onFailure: {
    githubStatusUpdate('failure')
  }
```

We need to do the equivalent if we add status updates here. Though it may not be as easy since those lifecycle hooks may not be as easily accessible

johnatan-velez: better to keep this in the dockerized pipeline, isn't it?
Commit: bb899139c6000350e02e37d4875b972faf40cea3

Jenkinsfile.rh.release
johnatan-velez: Commit: c5871aa8dd23af96fc332453aab1bec2f6e9de85

Jenkinsfile.rh.release
agustin-mrtz: I don't think so. We don't want to introduce the dockerized pipeline since that will build the image. Building the image happens in `build_and_push_rh_image.sh`. So I would say either find a way to give github status checks with out the dockerized pipeline or don't provide any at all (which may not be a good practice).

johnatan-velez: This could work: 2a67362a11cd2eb4e0733017dad7b6f27d60ded0

I only have one question whether I should throw the exception or not. @shaikhu  do you know? What do you think about this approach?

mr-zepol: > This could work: [2a67362](https://github.com/sonatype/docker-nexus-iq-server/commit/2a67362a11cd2eb4e0733017dad7b6f27d60ded0)
> 
> I only have one question whether I should throw the exception or not. @shaikhu do you know? What do you think about this approach?

@johnatan-velez Usman is out until September, so better check with someone else from BNR team

agustin-mrtz: Looks better now. I second Alex's advice of reaching out to BNR for feedback

Jenkinsfile.rh.release
```diff
@@ -24,9 +24,51 @@ properties([
   ]),
 ])
 
+String version = ''
+String checksum = ''
+
+void updateIQServerVersionAndChecksum(String version, String checksum) {
+  def dockerFile = readFile(file: 'Dockerfile.rh')
+  def metaShortVersionRegex = /(release=")(\d\.\d{1,3}\.\d)(" \\)/
+  def versionRegex = /(ARG IQ_SERVER_VERSION=)(\d\.\d{1,3}\.\d\-\d{2})/
```
kellyrob99: Not sure why this was ever done like this, but it's a complete anti-pattern. The entire point of build-args is to override them on the Docker build invocation. This would be much simpler if you just pass these in to your script invoking Docker and leave this file alone. 
https://docs.docker.com/engine/reference/commandline/build/#build-arg

Jenkinsfile.rh.release
```diff
@@ -24,9 +24,51 @@ properties([
...
+  def shaRegex = /(ARG IQ_SERVER_SHA256=)([A-Fa-f0-9]{64})/
+  dockerFile = dockerFile.replaceAll(metaShortVersionRegex,
+      "\$1${version.substring(0, version.indexOf('-'))}\$3")
+  dockerFile = dockerFile.replaceAll(versionRegex, "\$1${version}")
+  dockerFile = dockerFile.replaceAll(shaRegex, "\$1${checksum}")
+
+  writeFile(file: 'Dockerfile.rh', text: dockerFile)
+}
+
+void commitAndPushChanges(String version) {
```
kellyrob99: appears if you stop modifying Dockerfile.rh this also becomes unnecessary as part of releasing the image

Jenkinsfile.rh.release
johnatan-velez: Commit: a2a41fa10702ac5eaf527a8818f39a19baf2e028
Does this make more sense for you?
Would this replace the values for `IQ_SERVER_VERSION` and `IQ_SERVER_SHA256` wherever they appear in `Dockerfile.rh` ?

Jenkinsfile.rh.release
johnatan-velez: Commit: a2a41fa10702ac5eaf527a8818f39a19baf2e028

Jenkinsfile.rh.release
agustin-mrtz: +1 to Kelly's comment. It was inherited in other images to minimize the number of changes but in hindisght we should have gone this route which is what MTIQ and IQ (I think) do. We are also doing this on the NVS side of the house.

agustin-mrtz (approved):

PR #142, opened 2023-08-25
CLM-27310 Fix Red Hat Image Pipeline

agustin-mrtz:


johnatan-velez (approved):
+1

Jenkinsfile.rh.release
```diff
@@ -54,7 +58,7 @@ node('ubuntu-zion') {
             credentialsId: 'red-hat-api-token',
             variable: 'API_TOKEN')
       ]) {
-        sh 'PATH="$PATH:." VERSION=$version CHECKSUM=$checksum ./build_and_push_rh_image.sh'
+        sh 'PATH="$PATH:." VERSION=$version CHECKSUM=$checksum TAG=$tag ./build_and_push_rh_image.sh'
```
kellyrob99, line 61: VERSION here is no longer necessary? Isn't that also used as a buildArg for the Docker build? 
https://github.com/sonatype/docker-nexus-iq-server/blob/e43c67b44cc2e5b21f0b08d027a75641b1f708f3/build_and_push_rh_image.sh#L42

agustin-mrtz: I am not removing VERSION. I am just renaming it as tag since VERSION is referring to the IQ server version. I think I could re-name stuff to make it more descriptive

kellyrob99: But you have both of these variables on this line, and afaict `version` will now just be null.  Plus that variable is eventually used to download the IQ jar during the Docker build, isn't it? 
https://github.com/sonatype/docker-nexus-iq-server/blob/e43c67b44cc2e5b21f0b08d027a75641b1f708f3/Dockerfile.rh#L85C72-L85C89

agustin-mrtz: Version gets defined up above it won't be null 
```
version = getVersionFromBuildName(env.releaseBuild_NAME)
checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
```

agustin-mrtz: > Plus that variable is eventually used to download the IQ jar during the Docker build, isn't it?

Yes you are correct.

kellyrob99: tnx for the followup 👍

kellyrob99 (approved):

PR #144, opened 2023-08-31
Fix Red Hat Image IQ Jar Versioning

agustin-mrtz:


Jenkinsfile.rh.release
```diff
@@ -28,14 +28,12 @@ properties([
...
-String checksum = ''
-
 node('ubuntu-zion') {
   try {
     stage('Preparation') {
-      version = getVersionFromBuildName(env.releaseBuild_NAME)
-      checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
+      env.IQ_SERVER_VERSION = getVersionFromBuildName(env.releaseBuild_NAME)
+      env.IQ_SERVER_CHECKSUM = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${env.IQ_SERVER_VERSION}-bundle.tar.gz.sha256").trim()
+      env.IQ_RELEASE = env.IQ_SERVER_VERSION.split("-")[0]
```
johnatan-velez, line 36: Just to understand this because I have a limited understanding of these pipelines. Here you are accessing an environment variable or declaring it for first time? The question is because there is a `ARG IQ_RELESE` in the `Dockerfile.rh`. If they are different things, maybe use a different name would be helpful for new inexperienced modifiers.

johnatan-velez: Also, what's the advantage of using env.VARIABLE instead of the normal global variable?

agustin-mrtz: > Here you are accessing an environment variable or declaring it for first time? 

Declaring it for the first time

> The question is because there is a `ARG IQ_RELEASE` in the Dockerfile.rh. If they are different things, maybe use a different name would be helpful for new inexperienced modifiers. 

I added that argument. They refer to the same thing.

agustin-mrtz: > Also, what's the advantage of using env.VARIABLE instead of the normal global variable? 

From what I understood, the global variable is not accessible in the shell script. That is why we are encountering this bug. When the `sh` command runs the following occurs 
`PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:. VERSION= CHECKSUM= TAG=1.166.0-ubi-1 ./build_and_push_rh_image.sh`
Notice that `VERSION=` and `CHECKSUM=` are being set to nothing because the global variable is not accessible here

kakumara: > From what I understood, the global variable is not accessible in the shell script. That is why we are encountering this bug. When the `sh` command runs the following occurs
> `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:. VERSION= CHECKSUM= TAG=1.166.0-ubi-1 ./build_and_push_rh_image.sh`
> Notice that `VERSION=` and `CHECKSUM=` are being set to nothing because the global variable is not accessible here

This is a useful explanation and could easily be part of the PR description (for future reference)

kakumara: > From what I understood, the global variable is not accessible in the shell script. That is why we are encountering this bug.

We could test if this approach works by passing the env args to a test shell script and just printing (so that we don't have to wait till the next release to verify).

kakumara (approved):
+1, LGTM

johnatan-velez (approved):
+1 for the PR and also for the kasun's suggestion of testing with a test shell script.

Jenkinsfile.rh.release
kellyrob99: > > From what I understood, the global variable is not accessible in the shell script. That is why we are encountering this bug.
> 
> We could test if this approach works by passing the env args to a test shell script and just printing (so that we don't have to wait till the next release to verify).

1000% yes!

shaikhu (approved):

PR #158, opened 2024-07-15
Add add-opens flags

mattjohnson:
Some libraries use the reflection API to attempt to access non-public fields and methods of java.* APIs. This is no longer possible by default on JDK 17, but you can use the --add-opens option on the command line to enable it for specific tools and libraries. This adds the options to keep those libraries functioning as expected with the new Java 17 runtime.


langyizhao (approved):

jmromeroes (approved):
LGTM

PR #159, opened 2024-07-18
Bumps base image version

mattjohnson:


mattjohnson:
Using a different base image now.

PR #163, opened 2024-11-21
Uses assemblies in docker images and enables arm64+amd64 platforms

mattjohnson:
Changes included in this PR:
- Support both linux/amd64 and linux/arm64 architectures
- Update base images to use ubi-minimal
- Use assemblies that contain embedded JDK instead of using openjdk base image (includes too many unnecessary packages)
- Sets up docker trust for manifest list tags
- Builds all architectures on branch jobs

Overall docker image size is reduced by about 300MB.

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -29,15 +29,17 @@ properties([
 
 String imageName = 'sonatype/nexus-iq-server'
 String version = ''
-String checksum = ''
+String checksumX86_64 = ''
+String checksumAarch = ''
 
 dockerizedBuildPipeline(
   deployBranch: 'main',
   prepare: {
     githubStatusUpdate('pending')
     version = getVersionFromBuildName(env.releaseBuild_NAME)
-    checksum = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-bundle.tar.gz.sha256").trim()
-    updateIQServerVersionAndChecksum(version, checksum)
+    checksumX86_64 = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-linux-x86_64.tgz.sha256").trim()
+    checksumAarch = readBuildArtifact('insight/insight-brain/release', env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-linux-aarch_64.tgz.sha256").trim()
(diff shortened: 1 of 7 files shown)


Jenkinsfile.release
```diff
@@ -123,17 +121,16 @@ void pushImage(String imageName) {
...
-
-      withCredentials([string(credentialsId: 'sonatype-password', variable: 'DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE')]) {
-        runSafely "docker image push ${imageName}:${env.VERSION}"
-        runSafely "docker image push ${imageName}:latest"
+        sh  """
+        curl -L https://go.dev/dl/go1.23.3.linux-amd64.tar.gz | tar -xzf -
+        export PATH=${env.PATH}:${env.WORKSPACE}/go/bin:${env.WORKSPACE}/bin
+        export GOPATH=${env.WORKSPACE}
+        go install -tags pkcs11 github.com/theupdateframework/notary/cmd/notary@latest
+        notary --help
```
eduard-tita, line 130: `notary --help` - Is this line needed here?

Jenkinsfile.slim.release
```diff
@@ -117,17 +115,16 @@ void pushImage(String imageName) {
...
-
-      withCredentials([string(credentialsId: 'sonatype-password', variable: 'DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE')]) {
-        runSafely "docker image push ${imageName}:${env.VERSION}-slim"
-        runSafely "docker image push ${imageName}:latest-slim"
+        sh  """
+        curl -L https://go.dev/dl/go1.23.3.linux-amd64.tar.gz | tar -xzf -
+        export PATH=${env.PATH}:${env.WORKSPACE}/go/bin:${env.WORKSPACE}/bin
+        export GOPATH=${env.WORKSPACE}
+        go install -tags pkcs11 github.com/theupdateframework/notary/cmd/notary@latest
+        notary --help
```
eduard-tita, line 124: Same as above - is this line needed here?

eduard-tita (approved):
+0.5 Nothing stands out to me as wrong, but I cannot say I fully understand all the details. 

Especially the notarization approach is new to me, so I cannot tell by just looking at the PR if it works or not, or if anything is missing.

Jenkinsfile.release
mattjohnson: Not strictly needed. I liked it because it helped me debug locally to know it was installed and setup correctly.

PR #164, opened 2024-12-05
Tweak to minimize image size

kellyrob99:
Hey @mattjohnson I was admiring your work on https://github.com/sonatype/docker-nexus-iq-server/pull/163 to drastically reduce the image size - nice results using the builder!

I was doing some Docker image analysis today and noticed that the build ends up with two file layers related to the product directory - one on copy and again when the permissions are changes, effectively doubling up the size.
So this change subtracts another 270MB.

Smoke tested only, but I was able to sign into the app at the very least.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,21 +100,6 @@ RUN mkdir -p ${IQ_HOME} \
 && chmod 0755 ${CONFIG_HOME} \
 && chmod 0755 ${LOGS_HOME}
 
-# Copy config.yml
-COPY --from=builder /tmp/work/config-edited.yml ${CONFIG_HOME}/config.yml
-RUN chmod 0644 ${CONFIG_HOME}/config.yml
-
-# Copy server assemblies
-COPY --from=builder /tmp/work/nexus-iq-server ${IQ_HOME}
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -100,12 +100,21 @@ RUN mkdir -p ${IQ_HOME} \
...
+&& chown -R nexus:nexus ${CONFIG_HOME} \
+&& chown -R nexus:nexus ${LOGS_HOME}
+    
 # Copy config.yml
 COPY --from=builder /tmp/work/config-edited.yml ${CONFIG_HOME}/config.yml
 RUN chmod 0644 ${CONFIG_HOME}/config.yml
 
 # Copy server assemblies
-COPY --from=builder /tmp/work/nexus-iq-server ${IQ_HOME}
+COPY --chown=nexus:nexus --from=builder /tmp/work/nexus-iq-server ${IQ_HOME}
```
kellyrob99, line 117: Not entirely sure this is necessary, but I also don't see how it could hurt...

mattjohnson (approved):

mattjohnson:
Nice find!

PR #168, opened 2024-12-19
Updates RedHat image to match other images

mattjohnson:


jmromeroes (approved):
Basically the same that's already there on the other Docker files minus some specifics so makes sense to me

PR #170, opened 2025-03-21
Fix alpine image health check - CLM-34331

rpokorny:
https://sonatype.atlassian.net/browse/CLM-34331

`curl`, while included in the builder image, was not actually being included in the IQ server image, causing the healthcheck (which is implemented with curl) to always fail.


mealingr (approved):
+1

Qiming-C (approved):
+1

PR #171, opened 2025-04-16
BNR-1740: Slim Release

mburkert3:
JIRA: https://sonatype.atlassian.net/browse/BNR-1740

The `build_and_push_images.sh` script defaults to `Dockerfile` unless `DOCKERFILE` is explicitly set, causing the slim image to lag one version behind. Full findings and context are documented in the linked JIRA


shaikhu (approved):
Good catch!

mealingr (approved):
+1

PR #182, opened 2026-03-23
Remove most redhat docker images in favor of infosec hardened base images - CLM-39066

rpokorny:
https://sonatype.atlassian.net/browse/CLM-39066

This PR switches our default docker image to use Sonatype's own hardened base image instead of a RedHat base image.  This should make the image significantly lighter and cut down on policy violations coming from the containers.

For now, the other image builds will continue to exist, but the plan is to sunset most of them over the coming months in favor of only having this image.  The uncertain one is the actual "redhat" image which we use for Openshift - we must continue to support that as long as we support openshift, though that is also something we want to get away from.

Note that as part of this, I changed which IQ artifact the docker image builds upon.  In order to use the infosec-provided, hardened JRE, I needed to avoid using the jreleaser build of IQ.  Instead, the dockerfile now downloads the IQ server jar directly, along with a recently-added artifact specifying the JVM command-line flags with which the IQ server is supposed to be run.  It downloads these directly from sonatype.repo.sonatype.app using maven, instead of downloading from downloads.sonatype.com.  As such, I felt it was no longer necessary to hardcode the checksums to check into the Dockerfile.

Passing build (has to run against IQ snapshot until we do an IQ release with the changes from https://github.com/sonatype/insight-brain/pull/15541): https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/main-image/job/iq-server-docker-image-build-test/job/CLM-39066_hardened_base_images/36/


hanstdam:
Hi @rpokorny , 
I noticed the inclusion of a `Healthcheck.java` file. The HDS team had a similar approach at first, when we ran into memory issue of an additional java process being invoked. Therefore we've included a small health check utility called [`localcheck`](https://github.com/sonatype/localcheck) in all our base images, that I'd encourage you to use instead.

Jenkinsfile
```diff
@@ -30,9 +30,12 @@ void configureBranchJob() {
 
 String deployBranch = 'main'
 String imageName = 'sonatype/nexus-iq-server'
+String productionImage = 'iq-server-under-test'
 
 configureBranchJob()
 dockerizedBuildPipeline(
+  buildImageId: 'sonatype.repo.sonatype.app/docker-all/docker:latest',
```
rpokorny, line 27: Previously, our builds were (seemingly accidentally) using our own Dockerfile not only to build the output image, but to build _the build environment_ for that output image.  This was an unnecessary and inefficient complication.  So instead it will now use the plain `docker` image (e.g. an image that contains `docker` and can run our Dockerfile build) as the build image.

rpokorny:
> https://github.com/sonatype/docker-nexus-iq-server/pull/182#issuecomment-4158126608

Using localcheck as of [20b3405](https://github.com/sonatype/docker-nexus-iq-server/pull/182/commits/20b3405b393bd4b717df54fff1ba7174843733b7)

jmromeroes (approved):
LGTM, besides the conflicts that are showing up with main

SirMaster (approved):

Dockerfile
```diff
@@ -98,35 +102,27 @@ RUN mkdir -p ${IQ_HOME} \
...
 RUN echo "trap 'kill -TERM \`cut -f1 -d@ ${SONATYPE_WORK}/lock\`; timeout ${TIMEOUT} tail --pid=\`cut -f1 -d@ ${SONATYPE_WORK}/lock\` -f /dev/null' SIGTERM" > ${IQ_HOME}/start.sh \
-&& echo "/opt/sonatype/nexus-iq-server/bin/nexus-iq-server server ${CONFIG_HOME}/config.yml 2> ${LOGS_HOME}/stderr.log & " >> ${IQ_HOME}/start.sh \
+&& echo "java @${IQ_HOME}/jvm.options \$JAVA_OPTS -jar ${IQ_HOME}/nexus-iq-server.jar server ${CONFIG_HOME}/config.yml 2> ${LOGS_HOME}/stderr.log & " >> ${IQ_HOME}/start.sh \
 && echo "wait" >> ${IQ_HOME}/start.sh \
 && chmod 0755 ${IQ_HOME}/start.sh
 
 WORKDIR ${IQ_HOME}
 
-# enabling back support for SHA1 signed certificates
-RUN update-crypto-policies --set DEFAULT:SHA1
```
mealingr, line 118: Just looking at the commit which added this in https://github.com/sonatype/docker-nexus-iq-server/commit/5b14c0652dd087a2f91a66a84df2e0e06b8a7744

and wondering if this new base image already has SHA1 signed certificate support, or if it's obsolete now?

Dockerfile
```diff
@@ -14,53 +14,54 @@
...
-# hadolint ignore=DL3041,DL3040
-RUN mkdir -p ${TEMP} && \
-    microdnf update -y && \
-    microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git which crypto-policies crypto-policies-scripts
-
-# Copy config.yml and set sonatypeWork to the correct value
-COPY config.yml ${TEMP}
-
-# hadolint ignore=DL4006,SC3060
-RUN cat ${TEMP}/config.yml | sed -r "s/\s*sonatypeWork\s*:\s*\"?[-0-9a-zA-Z_/\\]+\"?/sonatypeWork: ${SONATYPE_WORK//\//\\/}/" > ${TEMP}/config-edited.yml
```
mealingr, line 25: Wondering about this part as well, it was replacing whatever `sonatypeWork` was set to with whatever the value of `$SONATYPE_WORK` was, but I'm guessing it's not needed anymore if `config.yml` already has the right value in

mealingr (approved):
+1

Dockerfile
rpokorny: Hmm nice catch.  I'm seeing that this was added for Azure database connections, and that Microsoft plans to have all of the Azure DB certs updated to a newer algo by "early 2026".  I'll need to do more research to confirm whether we still need to worry about this.

Dockerfile
rpokorny: Right yes, in this repo we supply our own config-yml, so it's silly to have to modify it on the fly to be right rather than just writing it the way we want in the first place.

Dockerfile
rpokorny: We should be safe here:
```
● Now I have both pieces of information. Let me summarize:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                         
  Complete Answers                                                                        
                                                                                                                                                                                                                                                                                         
  1. Azure Certificate Migration Timeline - EXACT DATES FOUND                                                                                                                                                                                                                            
   
  From the https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-security:                                                                                                                                                                                         
                                                                
  Intermediate certificate rotation schedule:                                                                                                                                                                                                                                            
  - West Central US and East Asia: Already complete             
  - UK South and US Government regions: Started January 21, 2026
  - Central US: Started January 26, 2026
  - All other regions: Started January 28, 2026                                                                                                                                                                                                                                          
   
  Root certificate rotation schedule:                                                                                                                                                                                                                                                    
  - China regions (DigiCert Global Root CA G1 → DigiCert Global Root G2): Started March 9, 2026
                                                                                                                                                                                                                                                                                         
  Since today is April 16, 2026, the Azure certificate migration is DONE or nearly done for almost all regions. The SHA1 certificates are being/have been replaced with SHA256.
                                                                                                                                                                                                                                                                                         
  2. Infosec JRE Image SHA1 Support - CONFIRMED                                                                                                                                                                                                                                          
   
  The infosec hardened JRE image (sonatype-infosec/jre:openjdk-17) has this configuration in java.security:                                                                                                                                                                              
                                                                
  jdk.certpath.disabledAlgorithms=MD2, MD5, SHA1 jdkCA & usage TLSServer, \
      SHA1 usage SignedJAR & denyAfter 2019-01-01                                                                                                                                                                                                                                        
   
  jdk.tls.disabledAlgorithms=SSLv3, TLSv1, TLSv1.1, DTLSv1.0, RC4, DES, \                                                                                                                                                                                                                
      ECDH, TLS_RSA_*, rsa_pkcs1_sha1 usage HandshakeSignature, \
      ecdsa_sha1 usage HandshakeSignature, dsa_sha1 usage HandshakeSignature
                                                                                                                                                                                                                                                                                         
  SHA1 is NOT completely blocked:
  - SHA1 for TLSServer (inbound TLS) is only blocked if signed by a JDK CA                                                                                                                                                                                                               
  - SHA1 for outbound TLS connections (like connecting to PostgreSQL) is NOT blocked
  - The rsa_pkcs1_sha1, ecdsa_sha1, dsa_sha1 algorithms are only blocked for handshake signatures
                                                                                                 
  Conclusion: The infosec JRE image allows SHA1 for database connections. The RHEL update-crypto-policies --set DEFAULT:SHA1 workaround was specifically because RHEL9's crypto policies were MORE restrictive than Java's defaults - they blocked SHA1 entirely. The Wolfi-based infosec
   image does not have this problem, so the workaround is unnecessary.
```

paulerickson:
@hanstdam is the intent of the base image that child images should all use the _same_ UID & GID for all files they handle?
It seems to me like it should ideally be passed into the specific image as an env var or argument by the user, so that they can have…
1. different UIDs per service (e.g. maybe I have `iq-prod`, `iq-uat`, `repo-prod`)
2. a way to **avoid** migrating files from one hardcoded UID to another, which could cause an extended outage IIUC

rpokorny:
> @hanstdam is the intent of the base image that child images should all use the same UID & GID for all files they handle?
> It seems to me like it should ideally be passed into the specific image as an env var or argument by the user, so that they can have…

> different UIDs per service (e.g. maybe I have iq-prod, iq-uat, repo-prod)
> a way to avoid migrating files from one hardcoded UID to another, which could cause an extended outage IIUC

I've done some research on this problem and from what I understand it is a very common yet utterly inconvenient problem with docker: how can we solve the problem of different UIDs per service while also retaining access to the files built into the image?  A runtime-dynamic UID won't have access to the files on the image.

According to my claude, the common solution to this is to pass in the desired UID as an env variable, and then _have the entrypoint be a shell script that starts as root and `chown`s everything before starting the app_.  What a hack! I'm shocked if that's really the common solution to this, and anyway I think it'd be a problem for us as we've had issues with slow-running `chown`s on startup before.  @paulerickson @hanstdam I'm very curious what you think.

paulerickson:
> I've done some research on this problem and from what I understand it is a very common yet utterly inconvenient problem with docker: how can we solve the problem of different UIDs per service while also retaining access to the files built into the image? A runtime-dynamic UID won't have access to the files on the image.
> 
> According to my claude, the common solution to this is to pass in the desired UID as an env variable, and then _have the entrypoint be a shell script that starts as root and `chown`s everything before starting the app_. What a hack! I'm shocked if that's really the common solution to this, and anyway I think it'd be a problem for us as we've had issues with slow-running `chown`s on startup before. @paulerickson @hanstdam I'm very curious what you think.

Not sure if it's the same thing as you're describing, but I've seen this approach from linuxserver.io & others, where they may use some arbitrary UID:GID like 1000:1000 at build time, but then at runtime the process looks for env vars `PUID` & `PGID` (the P is for parent) and changes its built-in user & group to those IDs (rather than `chown` everything in mounted volumes to match it's built-in IDs)

https://docs.linuxserver.io/general/understanding-puid-and-pgid/

https://github.com/linuxserver/docker-baseimage-ubuntu/blob/726a888ae20fd4065c2171b7aa90b3eb8408342d/root/etc/s6-overlay/s6-rc.d/init-adduser/run#L4-L15

I'm not sure if it's the _best_ way or not, but it would allow a user with some big existing sonatype-work dir to run with UID:GID matching the files they already have.

rpokorny:
> Not sure if it's the same thing as you're describing, but I've seen this approach from linuxserver.io & others, where they may use some arbitrary UID:GID like 1000:1000 at build time, but then at runtime the process looks for env vars PUID & PGID (the P is for parent) and changes its built-in user & group to those IDs (rather than chown everything in mounted volumes to match it's built-in IDs)

That is what I'm talking about, and I'm not sure how they did that and avoided the `chown`.  If the container changes the current user upon startup, then it's no longer the user that can access the files in the image, right?

hanstdam:
> is the intent of the base image that child images should all use the same UID & GID for all files they handle?

Yes. The fixed UID 65532 follows the industry standard (Google Distroless, Chainguard, Red Hat UBI all hardcode a single non-root UID). The intended way to override this per-service is via `securityContext.runAsUser` / `runAsGroup` in your pod specs, not in the image.

> a way to avoid migrating files from one hardcoded UID to another, which could cause an extended outage

This is a real concern. Since the previous base images used a different UID, customers moving to infosec-base-images will need to either update their `securityContext` / `fsGroup` to match 65532, or run a one-time `chown` via an `initContainer` on any persistent volumes. We should document this as part of the migration path.

> A runtime-dynamic UID won't have access to the files built into the image.

It will, as long as those files are readable (not just owner-readable). You don't need to own a file to read it. The only thing worth verifying is that our base images don't have files with overly restrictive permissions.

> the common solution to this is to pass in the desired UID as an env variable, and then have the entrypoint be a shell script that chowns everything before starting the app

> I've seen this approach from linuxserver.io & others, where they [...] change its built-in user & group to those IDs

Both patterns (mass chown, and the linuxserver.io PUID/PGID `usermod` approach) require starting the container as root and having a shell with user management utilities. That makes them incompatible with hardened/distroless images and read-only root filesystems. They originate from the Docker-on-a-NAS world (linuxserver.io's primary audience) where Kubernetes-level controls aren't available.

In a Kubernetes environment, the modern approach is to set file permissions correctly at build time (world-readable or group-readable for read-only content, group-writable for writable directories), and handle volume ownership via `fsGroup`. No runtime chown or usermod needed.

paulerickson:
Thanks for explaining @hanstdam!

Would the [--user](https://www.docker.com/blog/understanding-the-docker-user-instruction/) flag be the docker/podman CLI equivalent of what you're talking about in Kubernetes?
Are there any relevant links we can include in the readme?

hanstdam:
> Would the --user flag be the docker/podman CLI equivalent of what you're talking about in Kubernetes?

Yes, that's my understanding.

> Are there any relevant links we can include in the readme?

Here's what I've gathered:

**Handling volume ownership during UID migration:**
- Kubernetes: [Init Containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) for one-time `chown` before the app starts
- Kubernetes: [`fsGroup` in security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for automatic volume group ownership

**Overriding UID/GID at runtime (not in the image):**
- Docker/Podman: [`--user` flag](https://docs.docker.com/engine/containers/run/#user)
- Kubernetes: [`securityContext` with `runAsUser`, `runAsGroup`, `fsGroup`](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)
- ECS: [`user` parameter in container definitions](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_security) (same `uid:gid` format)

rpokorny:
@hanstdam is there a particular reason that it needs to be 65532 and not some other non-root UID? In looking at the user's migration path for this work, I see our old Dockerfile had its own non-root user: `nexus` with `UID` 1000.  If that's just as good from a security perspective (if not an industry-standards perspective), then I'd consider continuing to use that UID 1000 in this new Dockerfile, for the sake of backwards compatibility.

hanstdam:
> @hanstdam is there a particular reason that it needs to be 65532 and not some other non-root UID? In looking at the user's migration path for this work, I see our old Dockerfile had its own non-root user: `nexus` with `UID` 1000. If that's just as good from a security perspective (if not an industry-standards perspective), then I'd consider continuing to use that UID 1000 in this new Dockerfile, for the sake of backwards compatibility.

Hi @rpokorny,
We are simply following the model Chainguard is using. We're trying to create a base image that can be generally used by all our applications. The problem with changing to UID 1000, is that our [Nexus Repository images is using UID 200](https://github.com/sonatype/docker-nexus3/blob/main/Dockerfile#L60-L61), so if they start using our base images they'll have the same issue. The reality is that either we land on a organization-wide standard now, and deal with the migration, or we bloat the base images with multiple non-root UID. I would really like to avoid the latter.

hanstdam:
> > @hanstdam is there a particular reason that it needs to be 65532 and not some other non-root UID? In looking at the user's migration path for this work, I see our old Dockerfile had its own non-root user: `nexus` with `UID` 1000. If that's just as good from a security perspective (if not an industry-standards perspective), then I'd consider continuing to use that UID 1000 in this new Dockerfile, for the sake of backwards compatibility.
> 
> Hi @rpokorny, We are simply following the model Chainguard is using. We're trying to create a base image that can be generally used by all our applications. The problem with changing to UID 1000, is that our [Nexus Repository images is using UID 200](https://github.com/sonatype/docker-nexus3/blob/main/Dockerfile#L60-L61), so if they start using our base images they'll have the same issue. The reality is that either we land on a organization-wide standard now, and deal with the migration, or we bloat the base images with multiple non-root UID. I would really like to avoid the latter.

I misunderstood the question here. From a security perspective I can't think of any difference for using UID 1000 or 65532. As long as the user has non-root privileges.

rpokorny:
With ba2926a I have changed it back to use UID 1000 just like the redhat version did.  @hanstdam has confirmed in slack that this is just as secure and using the 65532 UID.  It's less compliant with industry standards, but is backwards compatible which I'd say is more important.

sonatype-iq:
### :smiley: All Clear! Sonatype Lifecycle didn't find any policy violations introduced by this PR
Well done. The committed code does not violate any of your organization's Sonatype Lifecycle policies.

----
### Sonatype Lifecycle Report Detail
**Application**: docker-nexus-iq-server
**Organization**: Lifecycle - IQ
**Date**: 2026-04-24 17:07:39 GMT
**PR Branch**: Develop Stage - [Full Report](https://sonatype.iq.sonatype.app/ui/links/application/docker-nexus-iq-server/report/d53d12d6b1b046da81c597e35a8c1893?source=pr-commenting)
**Base Branch**: Build Stage - [Full Report](https://sonatype.iq.sonatype.app/ui/links/application/docker-nexus-iq-server/report/08b4d2f951694de1a060b2d20d7e0879?source=pr-commenting)
**Application Priorities** - [View](https://sonatype.iq.sonatype.app/ui/links/developer/priorities/docker-nexus-iq-server/d53d12d6b1b046da81c597e35a8c1893)

[Give feedback](https://community.sonatype.com/t/user-feedback-github-pr-reviews/3811)

paulerickson (approved):
👍 except that README still has references to 65532, and looks like there is a merge conflict to resolve

rpokorny:
> 👍 except that README still has references to 65532, and looks like there is a merge conflict to resolve

8f91d9f

rpokorny:
It has been discovered that organizationally, we cannot currently use the infosec base images.

PR #183, opened 2026-04-17
CLM-39201: Remove shell from IQ Docker image

rpokorny:
https://sonatype.atlassian.net/browse/CLM-39201
https://sonatype.atlassian.net/browse/CLM-39570

## Summary

Removes busybox from the IQ Docker image by replacing the shell-based `start.sh` script with a C launcher. This reduces attack surface and aligns with the distroless intent of the infosec hardened base images.

## Changes

- **C launcher** (`launcher.c`): Handles stderr redirection and JAVA_OPTS parsing, then execs the JVM directly
- **Dockerfile**: Removes busybox, adds gcc to compile launcher, updates CMD to use launcher
- **expectations.groovy**: Rewritten to use host-side `/proc` inspection and `docker cp` instead of shell commands inside container
- **Jenkinsfile/Jenkinsfile.release**: Added JAVA_OPTS test flag for verification
- **README.md**: Documented shell removal as breaking change

## Breaking Changes

- The runtime image no longer contains a shell (`/bin/sh`)
- Customers who relied on `docker exec -it <container> sh` will need to use `docker cp` or ephemeral debug containers for troubleshooting

## Testing

Build and test locally, or rely on CI.

## Related

- Depends on CLM-39066 (infosec base image migration)
- Prerequisite for CLM-39207 (FIPS-compliant Docker image)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,9 +46,9 @@ RUN apk add --no-cache maven-3.9 openjdk-17-jre gcc \
 
 # Copy and compile the launcher with build-time paths
 COPY launcher.c /tmp/launcher.c
-RUN gcc -DIQ_HOME=${IQ_HOME} \
-         -DCONFIG_HOME=${CONFIG_HOME} \
-         -DLOGS_HOME=${LOGS_HOME} \
+RUN gcc -DIQ_HOME=\"${IQ_HOME}\" \
+         -DCONFIG_HOME=\"${CONFIG_HOME}\" \
+         -DLOGS_HOME=\"${LOGS_HOME}\" \
          -o /runtime-deps/bin/launcher /tmp/launcher.c
 
 # Create user/group entries in runtime root.
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ RUN apk add --no-cache maven-3.9 openjdk-17-jre gcc \
         --keys-dir /etc/apk/keys \
         --repositories-file /etc/apk/repositories \
         tini-static \
-        git
+        git-bootstrap
 
 # Copy and compile the launcher with build-time paths
 COPY launcher.c /tmp/launcher.c
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,9 +53,10 @@ RUN gcc -DIQ_HOME=\"${IQ_HOME}\" \
 
 # Create user/group entries in runtime root.
 # Include root user (UID 0) for system compatibility, plus nexus user.
+# Use nologin as shell since the runtime image has no shell anyway.
 RUN mkdir -p /runtime-deps/etc \
-    && echo "root:x:0:0:root:/root:/bin/sh" > /runtime-deps/etc/passwd \
-    && echo "nexus:x:${UID}:${GID}:Nexus IQ user:${IQ_HOME}:/bin/false" >> /runtime-deps/etc/passwd \
+    && echo "root:x:0:0:root:/root:/usr/sbin/nologin" > /runtime-deps/etc/passwd \
+    && echo "nexus:x:${UID}:${GID}:Nexus IQ user:${IQ_HOME}:/usr/sbin/nologin" >> /runtime-deps/etc/passwd \
     && echo "root:x:0:" > /runtime-deps/etc/group \
     && echo "nexus:x:${GID}:" >> /runtime-deps/etc/group
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,18 +24,21 @@ ARG IQ_SERVER_VERSION=1.202.0-01
 
 # Install Maven + JRE (for artifact download) and runtime deps into isolated root.
 # Runtime deps rationale:
-# - busybox: provides /bin/sh (runtime image is distroless, needs shell for start.sh)
-# - tini-static: init daemon for zombie process reaping
+# - tini-static: init daemon for zombie process reaping and signal forwarding
 # - git: required for IQ Server SCM integrations
+# - gcc: compile the launcher.c (build-time only, not copied to runtime)
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk
 # hadolint ignore=DL3018
-RUN apk add --no-cache maven-3.9 openjdk-17-jre \
+RUN apk add --no-cache maven-3.9 openjdk-17-jre gcc \
     && apk add --no-cache --initdb --root /runtime-deps \
         --keys-dir /etc/apk/keys \
         --repositories-file /etc/apk/repositories \
-        busybox \
         tini-static \
         git
+
+# Copy and compile the launcher
+COPY launcher.c /tmp/launcher.c
+RUN gcc -o /runtime-deps/bin/launcher /tmp/launcher.c
 
 # Download the server jar and JVM options file as individual Maven artifacts.
 # Uses BuildKit secret mount for settings.xml so credentials never appear in any layer.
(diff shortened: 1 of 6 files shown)


launcher.c
```diff
@@ -0,0 +1,94 @@
...
+#define CONFIG_HOME "/etc/nexus-iq-server"
+#endif
+#ifndef LOGS_HOME
+#define LOGS_HOME "/var/log/nexus-iq-server"
+#endif
+
+/*
+ * This launcher exists because the distroless runtime image has no shell.
+ * It replaces the previous shell-based start.sh and handles:
+ *   1. Redirecting stderr to LOGS_HOME/stderr.log
```
paulerickson, line 29: IMHO, it would be better _not_ to.  Do you know why this was done in the first place?  As a customer, I think I would expect stdout and stderr from the service to be output by the container.  If the process crashes, and wrote error output to its own ephemeral container storage, it would be harder for me to find.  Also, if it writes to ephemeral storage enough to fill it up, that could _cause_ the container to crash!

Dockerfile
```diff
@@ -18,25 +18,57 @@
...
+ARG GID=1000
+ARG UID=1000
 
 # Install Maven + JRE (for artifact download) and runtime deps into isolated root.
 # Runtime deps rationale:
-# - busybox: provides /bin/sh (runtime image is distroless, needs shell for start.sh)
-# - tini-static: init daemon for zombie process reaping
+# - tini-static: init daemon for zombie process reaping and signal forwarding
 # - git: required for IQ Server SCM integrations
+# - gcc: compile the launcher.c (build-time only, not copied to runtime)
```
paulerickson, line 27: How is it excluded from the runtime?

launcher.c
rpokorny: If this was a new container I would agree but at this point that would be a breaking change, which I'm seeking to avoid without good reason.

FWIW it writes to the same directory as all of the other log files, and should have little-to-no stderr output, so I'm not to worried about the storage/crashing concerns.

Dockerfile
rpokorny: Notice that there are two `apk` commands here: one installs normally into the builder image, and one installs into a `/runtime-deps` directory.  Only the latter gets copied to the runtime container.

launcher.c
paulerickson: Fair enough!

Dockerfile
```diff
@@ -18,25 +18,57 @@
...
+RUN gcc -DIQ_HOME=\"${IQ_HOME}\" \
+         -DCONFIG_HOME=\"${CONFIG_HOME}\" \
+         -DLOGS_HOME=\"${LOGS_HOME}\" \
+         -o /runtime-deps/bin/launcher /tmp/launcher.c
+
+# Create user/group entries in runtime root.
+# Include root user (UID 0) for system compatibility, plus nexus user.
+RUN mkdir -p /runtime-deps/etc \
+    && echo "root:x:0:0:root:/root:/bin/sh" > /runtime-deps/etc/passwd \
+    && echo "nexus:x:${UID}:${GID}:Nexus IQ user:${IQ_HOME}:/bin/false" >> /runtime-deps/etc/passwd \
```
paulerickson: If it's available, consider setting the `nologin` non-shell, which is commonly used for system users.  Root should not have a shell either, right?

```suggestion
    && echo "root:x:0:0:root:/root:/usr/sbin/nologin" > /runtime-deps/etc/passwd \
    && echo "nexus:x:${UID}:${GID}:Nexus IQ user:${IQ_HOME}:/usr/sbin/nologin" >> /runtime-deps/etc/passwd \
```

rpokorny: 🤖 Good suggestion! Fixed in fc4430f - using `/usr/sbin/nologin` for both root and nexus users since the runtime image has no shell anyway.

Dockerfile
paulerickson: Oh, I see now; thanks

paulerickson (approved):
👍 (not that I really know C, though 🤷)

mealingr (approved):
+1

launcher.c
```diff
@@ -0,0 +1,84 @@
...
+  // Redirect stderr to log file
+  int stderr_fd = open(LOGS_HOME "/stderr.log", O_WRONLY | O_CREAT | O_APPEND, 0644);
+  if (stderr_fd >= 0) {
+    dup2(stderr_fd, STDERR_FILENO);
+    close(stderr_fd);
+  }
+
+  // Build argument list
+  args[arg_count++] = "java";
+  args[arg_count++] = "@" IQ_HOME "/jvm.options";
```
mealingr, line 48: In the process of testing as part of https://sonatype.atlassian.net/browse/CLM-39621 and encountered this error

```
2026-04-27T17:53:26.896659814Z 2026-04-27 17:53:26,896+0000 WARN [SourceControlEventProcessor-0]  com.sonatype.nexus.git.utils.api.WithShellNativeGitUtils - Unable to determine if native git is available
2026-04-27T17:53:26.896683305Z org.zeroturnaround.exec.ProcessInitException: Could not execute [/bin/sh, -c, git --version]. Error=2, No such file or directory
```

Claude is suggesting we need to enable `-Dgit.without.shell=true` by default which was added as part of https://sonatype.atlassian.net/browse/CLM-31029

mealingr: > Claude is suggesting we need to enable -Dgit.without.shell=true

That env var didn't fix it for me using SSH I still got
```
2026-04-27T18:41:39.189682929Z Caused by: org.zeroturnaround.exec.InvalidExitValueException: Unexpected exit value: 128, allowed exit values: [0], executed command [git, clone, --depth, 1, --no-single-branch, --no-checkout, --branch, main, git@github.com:scm-test-2025/scm_test_maven.git, .] in directory /sonatype-work/source-control/9b06f20f6de4454ea83ce1ff0de24dbf with environment {GIT_TOKEN=<redacted>, SSH_ASKPASS=/sonatype-work/temp/nexus-iq-git-helper3245705139561450427.sh, GIT_TERMINAL_PROMPT=0, GIT_USERNAME=x-access-token, GIT_ASKPASS=/sonatype-work/temp/nexus-iq-git-helper3245705139561450427.sh, GIT_SSH_COMMAND=ssh -oBatchMode=yes}, output was 110 bytes:
2026-04-27T18:41:39.189686530Z Cloning into '.'...
2026-04-27T18:41:39.189687762Z fatal: cannot exec 'ssh -oBatchMode=yes': No such file or directory
2026-04-27T18:41:39.189689088Z fatal: unable to fork
```

mealingr: Same issue with "Use SSH for Git Operations" disabled
```
2026-04-28T10:24:27.496158623Z Caused by: org.zeroturnaround.exec.InvalidExitValueException: Unexpected exit value: 128, allowed exit values: [0], executed command [git, clone, --depth, 1, --no-single-branch, --no-checkout, --branch, main, https://github.com/scm-test-2025/scm_test_maven, .] in directory /sonatype-work/source-control/24a56169c7c94932aeb60308b682425a with environment {GIT_TOKEN=<redacted>, SSH_ASKPASS=/sonatype-work/temp/nexus-iq-git-helper17483636474646266520.sh, GIT_TERMINAL_PROMPT=0, GIT_USERNAME=x-access-token, GIT_ASKPASS=/sonatype-work/temp/nexus-iq-git-helper17483636474646266520.sh, GIT_SSH_COMMAND=ssh -oBatchMode=yes}, output was 214 bytes:
2026-04-28T10:24:27.496166993Z Cloning into '.'...
2026-04-28T10:24:27.496168433Z fatal: cannot exec '/sonatype-work/temp/nexus-iq-git-helper17483636474646266520.sh': No such file or directory
2026-04-28T10:24:27.496169513Z fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

rpokorny:
It has been discovered that organizationally, we cannot currently use the infosec base images.

PR #186, opened 2026-04-22
Skip git commit for patch releases when skipMarkAsLatest=true

jmromeroes:
## Summary
When `skipMarkAsLatest=true` (patch/hotfix release of older versions), the Dockerfile should NOT be updated in the main branch.

## Problem
The `commitAndPushChanges()` function in the Docker release Jenkinsfiles unconditionally commits and pushes the Dockerfile version updates to `main`, regardless of the `skipMarkAsLatest` parameter.

This caused the 1.199.1 patch release to incorrectly update the Dockerfiles from 1.202.1 (the latest) to 1.199.1 (the older patch version).

## Root Cause
- `skipMarkAsLatest` only affected whether to add the `latest` tag when pushing Docker images
- It did NOT skip the Dockerfile version update and commit to `main`

## Fix
Modified `commitAndPushChanges()` in all three affected Jenkinsfiles to:
1. Accept a `skipMarkAsLatest` parameter
2. Return early without making git changes when `skipMarkAsLatest=true`
3. Log why the commit is being skipped

## Affected Files
- `Jenkinsfile.release` (standard Docker image)
- `Jenkinsfile.slim.release` (slim Docker image)
- `Jenkinsfile.alpine.release` (alpine Docker image)

**Note:** `Jenkinsfile.rh.release` was not modified because it doesn't use `commitAndPushChanges()` and doesn't commit to the repository.

## Related
- Revert PR: #185 (reverts the incorrect version update)
- Confluence doc: https://sonatype.atlassian.net/wiki/spaces/SP/pages/2243264521 (Patch Release Process)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Jenkinsfile.alpine.release
+++ b/Jenkinsfile.alpine.release
@@ -42,7 +42,7 @@ dockerizedBuildPipeline(
       version = getVersionFromBuildName(env.releaseBuild_NAME)
       checksum = readBuildArtifact(params.sourceJob, env.releaseBuild_NUMBER, "artifacts/nexus-iq-server-${version}-linux_musl-x86_64.tgz.sha256").trim()
       updateIQServerVersionAndChecksum(version, checksum)
-      commitAndPushChanges(version)
+      commitAndPushChanges(version, params.skipMarkAsLatest)
     },
     pathToDockerfile: './Dockerfile.alpine',
     setVersion: {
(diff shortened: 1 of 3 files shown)


paulerickson (approved):
👍

PR #188, opened 2026-04-24
Proprietary license

rpokorny:
We recently made this repo private.  This accordingly updates its licenses not to be FOSS.


mealingr (approved):
+1

PR #190, opened 2026-04-28
Use Alpine without shell as default base image - CLM-39066

rpokorny:
After https://github.com/sonatype/docker-nexus-iq-server/pull/183 and https://github.com/sonatype/docker-nexus-iq-server/pull/182 got closed, this is the next attempt to move off of RedHat base images and remove the shell.  This time, using an alpine builder image and NO runtime base image.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "localcheck"]
+	path = localcheck
+	url = git@github.com:sonatype/localcheck.git
```
rpokorny, line 3: `localcheck` is a simple tool for http healthchecks written by our infosec dept.  At the moment, its only published form is a Wolfi-compatible apk package on our repo server, which we can't use in alpine. So instead the approach for the moment is to check out the source code as a submodule and build it locally during our docker build.

Dockerfile
```diff
@@ -14,60 +14,143 @@
...
+    && mkdir -p /runtime-deps/etc/nexus-iq-server \
+    && mkdir -p /runtime-deps/var/log/nexus-iq-server \
+    && mkdir -p /runtime-deps/tmp \
+    && chmod 1777 /runtime-deps/tmp \
+    && chown -R ${UID}:${GID} /runtime-deps/opt/sonatype/nexus-iq-server \
+    && chown -R ${UID}:${GID} /runtime-deps/sonatype-work \
+    && chown -R ${UID}:${GID} /runtime-deps/etc/nexus-iq-server \
+    && chown -R ${UID}:${GID} /runtime-deps/var/log/nexus-iq-server
+
+# Download the server jar and JVM options file as individual Maven artifacts.
```
rpokorny, line 113: An alternative that I considered, since we aren't using infosec's specially hardened JVM in this approach, was to download the jreleaser bundles and not have a globally-installed JRE on the docker image.  However, that won't work because we don't have a jreleaser bundle for aarch64 MUSL, and apparently the reason we don't is because temurin doesn't provide a build for that configuration.  So system JVM it is.

Jenkinsfile
```diff
@@ -30,9 +30,12 @@ void configureBranchJob() {
 
 String deployBranch = 'main'
 String imageName = 'sonatype/nexus-iq-server'
+String productionImage = 'iq-server-under-test'
 
 configureBranchJob()
 dockerizedBuildPipeline(
+  buildImageId: 'sonatype.repo.sonatype.app/docker-all/docker:latest',
```
tneer, line 37: I'm never a fan of using `latest` even for things like this.  A major version change may lead to breaking changes and it's hard to track down.

I'm also a little surprised this works given the struggles we had with docker in docker on Jenkins.  I couldn't find a job that ran this Jenkinsfile.    I'm presuming it is https://jenkins.ci.sonatype.dev/job/insight/job/insight-brain/job/docker/job/main-image/job/iq-server-docker-image-build-test/job/CLM-39066_alpine_default_base/ which hasn't been run.

Jenkinsfile
```diff
@@ -42,33 +45,63 @@ dockerizedBuildPipeline(
     hadolint(['./Dockerfile'])
   },
   buildAndTest: {
-    def expectations = load 'expectations.groovy'
-    validateExpectations(expectations.containerExpectations())
+    withSonatypeDockerRegistry() {
+      sh 'echo $JENKINS_DOCKER_PASSWORD | docker login -u $JENKINS_DOCKER_USERNAME --password-stdin sonatype.repo.sonatype.app'
```
tneer, line 49: `docker login` leaks credentials.  Maybe they are in the container/workspace that gets cleaned up rather than a shared location on the agent but we should confirm.

PR #193, opened 2026-06-04
chore: remove non-essential vulnerable packages from final image

rpokorny:
## Summary

Remove `rsync` from the final runtime image. IQ Server never invokes rsync — no code path in `NativeGitCommands` or any other runtime component uses it. This eliminates [CVE-2026-41035](https://guide.sonatype.com/vulnerability/CVE-2026-41035).

## Why only rsync?

Other vulnerable transitive dependencies (gnutls, libtasn1, libxml2, sqlite-libs) were evaluated for removal but **must be retained** because they have shared library consumers that remain in the image:

- **libtasn1** — `p11-kit` links against `libtasn1.so.6` for ASN.1 certificate parsing. Removing it breaks TLS cert validation.
- **gnutls** — `systemd-libs` may link against `libgnutls.so.30`. Cascading failures possible.
- **libxml2** — May be linked by systemd, dbus-libs, or other base image components.
- **sqlite-libs** — `nss-softokn` links against `libsqlite3.so.0`. Removing it could break the certificate trust chain.

Removing these with `rpm -e --nodeps` would risk silent runtime failures that our test suite wouldn't catch.

## Testing

- [ ] CI build passes (all three Dockerfile variants)
- [ ] IQ scan confirms rsync CVE is gone

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ ARG SONATYPE_WORK="/sonatype-work"
 # hadolint ignore=DL3041,DL3040
 RUN mkdir -p ${TEMP} && \
     microdnf update -y && \
-    microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y procps gzip unzip tar shadow-utils findutils util-linux less rsync git-core openssh-clients which crypto-policies crypto-policies-scripts
+    microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y gzip unzip tar shadow-utils findutils less rsync git-core openssh-clients which crypto-policies crypto-policies-scripts
 
 # Copy config.yml and set sonatypeWork to the correct value
 COPY config.yml ${TEMP}
@@ -88,7 +88,7 @@ USER root
 # For testing
 # hadolint ignore=DL3041
 RUN microdnf update -y \
-&& microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y procps gzip unzip shadow-utils findutils util-linux less rsync git-core openssh-clients which crypto-policies crypto-policies-scripts \
+&& microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y gzip unzip shadow-utils findutils less git-core openssh-clients which crypto-policies crypto-policies-scripts \
 && microdnf clean all
 
 # Create folders & set permissions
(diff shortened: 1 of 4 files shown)


JuanDavid31:
## `--nodeps` verification

The `rpm -e --nodeps` forces removal ignoring dependency checks. Before merging, can you confirm these packages have no runtime dependents?

```bash
rpm -q --whatrequires gnutls libtasn1 libxml2 sqlite-libs
```

If only build-time tools require them, forced removal is safe. If something like `curl` or `git-core` requires `gnutls` or `libxml2`, removing with `--nodeps` could cause runtime failures.

Just want to ensure we've verified the dependency graph before forcing these out.

j-s-3 (approved):

rpokorny:
🤖 > The `rpm -e --nodeps` forces removal ignoring dependency checks. Before merging, can you confirm these packages have no runtime dependents?

Good call — we dug into this thoroughly. The original approach of `rpm -e --nodeps gnutls libtasn1 libxml2 sqlite-libs` was indeed risky. After investigation:

**libtasn1 cannot be removed** — `p11-kit-trust.so` links directly against `libtasn1.so.6` (confirmed via `ldd` in the container). Removing it would break TLS certificate validation.

**gnutls, libxml2, sqlite-libs CAN be removed** — but not with `--nodeps`. We verified via `ldd` in a built container that no runtime binary (git, ssh, curl) links against any of these. Their only RPM-level dependents were packages in the package-management stack (libarchive, gnupg2, etc.) which are also being removed.

The PR has been reworked. The new approach:
1. `rpm -e --nodeps` is used only for 3 packages with phantom RPM deps that aren't actual .so links: `gawk` (krb5-libs scriptlet dep), `libfido2` (openssh-clients declares dep but `/usr/bin/ssh` doesn't link against it — only `ssh-sk-helper` does), and `systemd-libs` (only libfido2 depended on it).
2. Everything else — including gnutls, libxml2, sqlite-libs — is removed via `microdnf remove` with proper dependency resolution (24 packages cleanly resolved).

Also removed `procps-ng` and `util-linux` from the install list entirely — neither is used by IQ Server at runtime, and removing them eliminates the systemd-libs dependency.

Verified end-to-end in a container build: git, ssh, curl all functional; TLS cert validation confirmed via `git ls-remote https://github.com/sonatype/docker-nexus-iq-server.git HEAD`.

rpokorny:
🤖 **Update to my previous comment:** After further investigation, `libtasn1` (and `p11-kit`/`p11-kit-trust`) CAN in fact be removed too.

My earlier statement that "libtasn1 cannot be removed because p11-kit-trust.so links against it" was technically true about the `.so` link — but it turns out nothing at runtime ever *loads* `p11-kit-trust.so`. Verified via:
- `LD_DEBUG=libs` during TLS operations shows libtasn1 is never loaded
- No binary or `.so` in the image has an `ldd` dependency on `libp11-kit.so.0`, `p11-kit-trust.so`, or `libnssckbi.so`
- OpenSSL reads the PEM trust bundle (`/etc/pki/tls/certs/ca-bundle.crt`) as a flat file — it does not use the PKCS#11 interface

These modules are only used at **build time** by `update-ca-trust` (called by `update-crypto-policies`) to generate the PEM bundle. Once the bundle exists on disk, they are unused.

The latest push adds `p11-kit`, `p11-kit-trust`, and `libtasn1` to the `rpm -e --nodeps` line (they need `--nodeps` because `ca-certificates` declares a dep on `p11-kit-trust`, but that dep is for trust store *generation*, not runtime *consumption*). TLS cert validation confirmed working after removal via `git ls-remote` over HTTPS.

j-s-3 (approved):

JuanDavid31 (approved):

